PR #13, opened 2021-10-01
feat: Only supply TypeScript types

achou11:
BREAKING CHANGE: Completely removes Flow types and schema validators


gmaclennan (commented):
Looks good to me, although I haven't checked the types in detail. Maybe you could test you can import it ok in Mapeo Mobile (to ensure that the package.json fields are picked up by Typescript correctly) by doing `npm pack` in the mapeo-schema folder, then `npm install $path_to_npm_pack_tarball` in the mapeo-mobile folder

achou11:
> Looks good to me, although I haven't checked the types in detail. Maybe you could test you can import it ok in Mapeo Mobile (to ensure that the package.json fields are picked up by Typescript correctly) by doing `npm pack` in the mapeo-schema folder, then `npm install $path_to_npm_pack_tarball` in the mapeo-mobile folder

Just tried this and looks like it's reading it properly 👍

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+        }[];
+        created_at: string;
+        deviceId?: string;
+        id: string;
+        lat?: number | null;
+        links?: string[];
+        lon?: number | null;
+        metadata?: {
+            location?: {
+                error: boolean;
```
achou11, line 32: Is this helpful to keep? doesn't seem like it's used by any client-side consumer

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+                permission: "granted" | "denied" | "never_ask_again";
+                position?: Position;
+                provider?: {
+                    backgroundModeEnabled: boolean;
+                    gpsAvailable: boolean;
+                    passiveAvailable: boolean;
+                    locationServicesEnabled: boolean;
+                    networkAvailable: boolean;
+                };
```
achou11, line 41: do we want this to reflect the same exact thing as what Expo defines? (this is how we get this info)

https://github.com/expo/expo/blob/a71ca98f1db974e2c99d04c305c94ae6068ee6a9/packages/expo-location/src/Location.types.ts#L141-L150

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+            accuracy?: number;
+            altitude?: number;
+            heading?: number;
+            latitude?: number;
+            longitude?: number;
+            speed?: number;
+        };
+        mocked?: boolean;
+        timestamp: number;
+    }
```
achou11, line 17: do we want this to be the same exact thing as what Expo defines?

https://github.com/expo/expo/blob/a71ca98f1db974e2c99d04c305c94ae6068ee6a9/packages/expo-location/src/Location.types.ts#L120-L134

index.d.ts
ErikSin: How did you come to this conclusion?

achou11: i only checked for mobile and desktop, but did a code search for anything trying to access `location.error` and didn't find any instances 🤷

ErikSin:
Would it be useful to document where this schema is being used? Like specific files in desktop and mobile

achou11:
> Would it be useful to document where this schema is being used? Like specific files in desktop and mobile

Personally don't think it would be (or at least as detailed as the file-level). If we listed projects using it then maybe, although the goal is that any project working with mapeo data should be adhering/using this

index.d.ts
ErikSin: I see! No, that makes a lot of sense. I guess this brings up the next question, is this is being used anywhere else?

achou11: you mean like external projects? this PR is a breaking change regardless and we'll create a new major version for these changes. if any external user relies on the definitions from the previous major version, they'll just stick with that.

Gregor did mention that at least for this specific field, it might be useful to have, but like I said it doesn't seem to be used anywhere our current projects

ErikSin:
> > Would it be useful to document where this schema is being used? Like specific files in desktop and mobile
> 
> Personally don't think it would be (or at least as detailed as the file-level). If we listed projects using it then maybe, although the goal is that any project working with mapeo data should be adhering/using this

That make sense. I think the question stems from a place of feeling like we are working backwards trying to create definitions (obviously this was created AFTER mobile and desktop). Right now I'm just trying to wrap my head around why we are using these definitions, so I will dive deeper into mobile and desktop and work from there

ErikSin (approved):
Perhaps we need to have a bigger conversation about how we are using these types, but this does seem like a good start off point. @gmaclennan I think it would still be good to get your eyes on Andrews questions

index.d.ts
ErikSin: @gmaclennan

index.d.ts
ErikSin: @gmaclennan I think it can't hurt to move towards this, but gregor you have worked more closely with this. Do you have any thoughts

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+            };
+            manualLocation?: boolean;
+        };
+        refs?: {
+            id: string;
+        }[];
+        schemaVersion: 3;
+        tags?: {
+            [key: string]: any;
+        };
```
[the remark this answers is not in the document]
gmaclennan: In Desktop we support any JSON value here. On mobile I don't _think_ we support as much, but I think there is a fallthrough on the code if it's not a string or number? We definitely support arrays for multi-selects. Mobile will only show fields that have a preset defined I think. Desktop will do some heuristics to try to figure out the best way to display any value.
TL;DR I would say [JSONValue](https://github.com/sindresorhus/type-fest#basic) if mobile will gracefully fail on stuff it cannot display.

index.d.ts
gmaclennan: Yes I think since we get this from Expo, the type should be what Expo says. Not sure where `mocked?` comes from or whether we need it?

index.d.ts
```diff
@@ -0,0 +1,135 @@
...
+        created_at: string;
+        deviceId?: string;
+        id: string;
+        lat?: number | null;
+        links?: string[];
+        lon?: number | null;
+        metadata?: {
+            location?: {
+                error: boolean;
+                permission: "granted" | "denied" | "never_ask_again";
```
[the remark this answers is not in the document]
gmaclennan: Yeah I think this is only here because I lazily wrote the whole of the location context on mobile to this value, and I don't think this should have ever been written to values. I think we can drop it from the type definition, and not assume it exists in any code we write.
However... thinking about it a bit more, I guess there is a case where this could be useful for troubleshooting, to figure out _why_ a particular point does not have a location. Hmmm... thinking out loud, but maybe worth keeping?

index.d.ts
achou11: right, but for the sake of this initial update, will use the same definition as what Expo currently has

gmaclennan: Right yes, sorry I had not noticed that 3 of the properties are optional in the expo definition. Let's use that.

gmaclennan:
Mobile and desktop started off with slightly different types for this data, which caused problems for compatibility. I tried to consolidate everything into this repo, but did not completely transition mobile and desktop to use this, because Desktop was not using Flow for a while, and a lot of the mobile components expected the data to be transformed at runtime to match the previous type. It's really important that the type definition for data that we write to Mapeo Core is clearly defined in one location, because if data is written that does not match this type, then it can break client code (e.g. the client could read data that is missing an expected value). One of the tasks I would like to do in Mapeo Core is to validate the data (and transform old data with a different schema) before sending it to the client, so that the clients (mobile and desktop) have strong type guarantees on the observations and other records they read from Mapeo Core.

achou11:
closing in favor of #16 lots has changed since I worked on this and think @sethvincent has a better idea of what this package should be moving forward

PR #16, opened 2022-09-13
switch to typescript types, update deps

sethvincent:
This PR switches from flow to typescript.

There might be remaining work around the examples to make sure they still work.


sethvincent:
closing in favor of #22

PR #22, opened 2022-11-29
feat: protocol buffers, json schema, & types for each

tomasciccola:
This PR implements a migration to using protocol buffers as encoding scheme for records in mapeo. 
Typescript types are generated from a .proto file for each possible record type (this PR only implements Observation).

Additionally, it exposes an encode and decode functions that validate the records against a json schema.


index.js
```diff
@@ -0,0 +1,78 @@
...
+import addFormats from 'ajv-formats'
+
+const loadJSON = (path) =>
+  JSON.parse(fs.readFileSync(new URL(path, import.meta.url)))
+const ajv = new Ajv()
+addFormats(ajv)
+
+const schemaTypesMap = {
+  observation: {
+    magicByte: 0, // this is one byte-length (0-255)
```
sethvincent, line 18: I've been calling this `blockPrefix` in mapeo-core-next rather than `magicByte`. It probably doesn't matter too much what we end up with.

I realized reading through this though that having one byte is perhaps not enough if we're also indicating the version of the schema. Maybe two: first byte to indicate schema and second byte to indicate version?

proto/observation.proto
```diff
@@ -0,0 +1,69 @@
+syntax = "proto3";
+
+package observation;
+import "google/protobuf/struct.proto";
```
sethvincent, line 4: Reading this I _just_ learned that we can possibly use this import to do the following instead of including Value and Struct definitions in this file:

```
google.protobuf.Value
google.protobuf.Struct
```

That would clean this up a bit and make it just about our data model, which would be nice.

index.js
tomasciccola: `blockPrefix` does a better job illustrating the use for that than `magicByte`, so I'll probably change it.

index.js
```diff
@@ -0,0 +1,84 @@
...
+const schemaTypesMap = {
+  observation: {
+    blockPrefix: 'obse', // 4 byte string
+    protobufSchema: Observation,
+    jsonSchema: loadJSON('./schema/observation.json'),
+  },
+}
+/**
+ * Encode a an object validated against a schema as a binary protobuf to send to an hypercore.
+ * @param {import('./types/schema/observation')} obj - Object to be encoded
```
gmaclennan, line 26:
```suggestion
 * @param {import('./types/schema/observation').Observation} obj - Object to be encoded
```

index.js
```diff
@@ -0,0 +1,84 @@
...
+ * @returns {Buffer} protobuf encoded buffer with 2 bytes prepended, one for the type of record and the other for the version of the schema */
+export const encode = (obj) => {
+  const record = Object.assign({}, obj)
+  const recordType = schemaTypesMap[record.type]
+  const validate = ajv.compile(recordType.jsonSchema)
+  const isValid = validate(record)
+  assert(isValid, JSON.stringify(validate.errors, true, 2))
```
gmaclennan, line 33:
```suggestion
  assert(isValid, JSON.stringify(validate.errors, null, 2))
```
putting `true` as the second arg is invalid.

index.js
```diff
@@ -0,0 +1,84 @@
...
+    ...record,
+    id: record.id.toString('hex'),
+    type: recordType,
+    schemaVersion: parseInt(schemaVersion),
+    version: `${opts.key.toString('hex')}/${opts.index.toString()}`,
+  }
+
+  const validate = ajv.compile(schemaTypesMap[recordType].jsonSchema)
+  const isValid = validate(record)
+  assert(isValid, JSON.stringify(validate.errors, true, 2))
```
gmaclennan, line 81:
```suggestion
  assert(isValid, JSON.stringify(validate.errors, null, 2))
```

gmaclennan (commented):
Sorry for the delay reviewing this. Every time I looked at it I had trouble getting `buf` installed on my machine but it is working now (it kept wanting to download some 20GB XCode update but I worked around it).

Good work on this, this is good progress. I haven't had time to review anything but I don't want to hold you up more, so here are some quick thoughts/feedback:

- Tests: I don't think we need Hypercore in there, it's not relevant to what is being tested.
- Tests: This will be more maintainable if you create a folders of fixtures, e.g. valid docs, and invalid docs, tests can just then iterate through these. To check encoding works I think we also need to decode and check equality.
- blockPrefix: @sethvincent I remember when we discussed we had talked about a unique ID, to enable ecosystem growth without needing to register types. It has a small size overhead, depending on the size of the prefix we use. 32 bytes would definitely avoid collisions, but maybe overkill. 16 bytes would be probably enough? e.g. each schema type is a randomly generated 16 byte Buffer.
- SchemaVersion: 1-byte (255 schema versions), feels like enough, but just want to check we are ok with this and won't be kicking ourselves later. Maybe 2-bytes to be really sure?
- Terminology: Can we think of some names / vocabulary to refer to the data type that is encoded to protobuf vs. the data type that is returned from the `decode()` function (e.g. it's the difference between the type from the protobuf and the type from the json schema). I feel like in the code it's confusing, and it's hard to talk about without clear terminology.
- Validation: I don't think we need to validate in the decode function, because doesn't the protobuf decode guarantee the type / validate the doc?
- Validation: I'm not sure validation should be part of `encode`, for performance sake? I think maybe this module should export the validation function, then the consumer of this API can choose when to use it (e.g. it should be used when the observation is coming from unknown input, but when it is being passed around as a known type then the validation is not needed)
- Validation: validate functions should not be compiled in the function body each time, for performance reasons.
- Assertions: I don't think assertions are necessary if we are using typescript / writing type safe code.
- Types: The index.js code currently fails type checks. I can help with fixing that if needed.
- package.json main field is incorrect.
- For performance, it's maybe best for the schemaTypesMap to be indexed by blockPrefix? (as a hex-encoded string maybe if we are going for random types).
- Maybe keep the block prefix / record type ID alongside the schema somehow? It seems very closely related to it. I can discuss ideas if needed.

I have a bunch of questions / comments about both protobuf encoding and the json schema, but maybe we can leave that outside the scope of this PR?

tomasciccola:
Thanks for the comments! I'll try to address them individually as I work through them:
* Tests: ok, I'll remove hypercore and add testing equality from encoded and decoded records
* blockPrefix: yeah, I was under the intuition that you wanted to allow third parties to add new records without registering them.
* schemaVersion: yeah, I think 2 bytes just to be sure.
* terminology: ok. I'll think of a better naming scheme; I agree its kinda confusing...
* validation on decoding: yeah, you're right. Validation on decoding is kinda unnecessary
* validation on encoding: ok. I'll just expose a validation function and compile each validation fn behorehand for each record type.
* assertion: ok, removing them!
* if I ran `tsc --allowJs index.js` it doesn't throw any error (just complaining that it can't transpile the file since its already js...)
* I agree blockPrefix should be close to the jsonSchema as a way of documenting them, instead of having it inside the module as an object. I'm gonna think about an approach that doesn't involve the consumer knowing about the prefixes.

tomasciccola:
Heh, as soon as I added //@ts-check on the file I started seeing my editor reporting those type errors :+1:

gmaclennan:
Suggestion: For the data type ID and the schema version, we use the JSON schema [`$id` property](https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1).

This must be a URI, but does not have to actually point to anything. It could be:

```
https://mapeo.app/schemas/:dataTypeId/:schemaVersion
```

That way we can parse it to extract the relevant data. We also need to define the "friendly" name for the data type (e.g. "observation"). This could just be the filename, or we could encode it in the $id, or we could store everything in a sidecare "metadata" file, or stick it somewhere else in the JSON schema.

We also need a similar way of putting this alongside the protobuf schemas.

I'm imagining that with this data alongside the schema files then the main code in `index.js` does not need to hard-code the schema types (e.g. what is currently in the `blockPrefixToSchema` variable), but can somehow generate that from the files themselves.

gmaclennan:
> We also need to define the "friendly" name for the data type (e.g. "observation").

I'm realizing that this is already stored in the `type` property of the JSON schema, and we don't need it in the protobuf, so we can just use that.

tomasciccola:
> or the data type ID and the schema version, we use the JSON schema [`$id` property](https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1).

The issue I'm finding with this approach is when decoding a record. If you parse the first 32 bytes you'll get the dataTypeId. But then, to know to which recordType it belongs one need to go through each schema and parse the $id to see if it corresponds to that recordType.
We'll always have this issue since we go from `friendlyName -> dataTypeId` when encoding and vise versa when decoding.
The approach I was using was having two maps in index.js: `recordTypeToBlockPrefix` and `blockPrefixToSchema`. I don't like that much this approach and I still have to go through each record when decoding (index.js:88) so I dunno...

tomasciccola:
I mean, we can use the $id URL approach to save the canonical source of info regarding dataTypeId, but I would still have a map that uses that for better lookup. 
And I think we don't need to put the version on the $id since its already a field on the schema (and it also allow to force a version)

gmaclennan:
> > or the data type ID and the schema version, we use the JSON schema [`$id` property](https://json-schema.org/draft/2020-12/json-schema-core.html#section-8.2.1).
> 
> The issue I'm finding with this approach is when decoding a record. If you parse the first 32 bytes you'll get the dataTypeId. But then, to know to which recordType it belongs one need to go through each schema and parse the $id to see if it corresponds to that recordType. We'll always have this issue since we go from `friendlyName -> dataTypeId` when encoding and vise versa when decoding. The approach I was using was having two maps in index.js: `recordTypeToBlockPrefix` and `blockPrefixToSchema`. I don't like that much this approach and I still have to go through each record when decoding (index.js:88) so I dunno...

Yes I still think it makes sense to have the maps to speed up lookup. The reason for my suggestion was to keep the data type ID alongside the schema definition, to avoid needing to update two places and keep them in sync as we add new data types. I was imagining something like (pseudo-code):

```js
const schemaFiles = await glob('schemas/*.json')
const blockPrefixToSchema = {}
for (const file of schemaFiles) {
  const schema = await loadJSON(file)
  const { schemaVersion, dataTypeId } = parseId(schema.$id)
  const type = schema.properties.type.enum[0]
  blockPrefixToSchema[dataTypeId + schemaVersion] = { type , schema, validate: ajv.compile(schema), protobufSchema: await import(`./types/proto/${type}.js`)
}
```

However might be better to generate those maps in JSON as a build step to avoid the need for runtime fs reads and dynamic imports... But I realize this is just a perf optimization and a way of reducing maintenance in the future, so I don't think it's critical right now.

tomasciccola:
> Perhaps something in the filename or folder names? We should also consider schema version in the filename / folder name: as we evolve schemas we would want to add the new schema to this repo while keeping the previous version — and in the future we will probably want to write something to convert an older schema version of a document to the latest.

Continuing with this, if we're going to keep various schema versions alongside, do we version each schema independently or all together?
If we version them together we can have a folder structure like 
`schema/{v1,v2}/{observation, common}.json`
if we version each independently we could do smth like
`shema/{observation,common}/{v1,v2}.json`
wdyt?

gmaclennan:
> > Perhaps something in the filename or folder names? We should also consider schema version in the filename / folder name: as we evolve schemas we would want to add the new schema to this repo while keeping the previous version — and in the future we will probably want to write something to convert an older schema version of a document to the latest.
> 
> Continuing with this, if we're going to keep various schema versions alongside, do we version each schema independently or all together? If we version them together we can have a folder structure like `schema/{v1,v2}/{observation, common}.json` if we version each independently we could do smth like `shema/{observation,common}/{v1,v2}.json` wdyt?

Independently, because I think it's likely that we only need to change some of them. E.g. we might end up with 6 data types in here, and then for a feature we realise that we need to change a property on just one of them, but the other datatypes remain the same.

tomasciccola:
> However might be better to generate those maps in JSON as a build step to avoid the need for runtime fs reads and dynamic imports... But I realize this is just a perf optimization and a way of reducing maintenance in the future, so I don't think it's critical right now.

the issue with this is that we can't serialize the protobufSchema encode and decode methods nor the validate fns

gmaclennan:
> the issue with this is that we can't serialize the protobufSchema encode and decode methods nor the validate fns

Yes, true, you would need to build the full map at runtime, e.g. append the encode and decode methods. But for now maybe just write the map manually and we can just update it when we add other schemas.

index.js
```diff
@@ -0,0 +1,147 @@
...
+    blockPrefix,
+    `schema of type ${obj.type} and version ${obj.schemaVersion} not found`
+  )
+  const schema = blockPrefixToSchema[blockPrefix]
+  // id is a hex encoded string, but is turned into bytes when protobufed,
+  // so we turn it into a buffer before that. TS doesn't like this since we are turning a string (id) into a buffer in-place
+  record.id = b4a.from(record.id, 'hex')
+  // how can we crash if blockPrefix is undefined? shouldn't an assertion be enough?
+  const dataTypeId = b4a.from(blockPrefix.split('/')[0])
+  const version = b4a.from([record.schemaVersion])
```
gmaclennan, line 116: This will not work if schema version > 255, or if schema version size is > 1.
Look at [compact-encoding](https://github.com/compact-encoding/compact-encoding) for some ideas about how to encode this.

gmaclennan (changes requested):
Great work on this, there are a lot of moving pieces for this kind of schema work!

Unfortunately, we need some build steps and code generation for this to actually work in the app. We bundle our code, so any `fs.readFile()`, `glob()` and dynamic import (`await import()`) will not work at runtime in the app.

I also think it's worth generating [standalone validation code](https://ajv.js.org/standalone.html) from AJV.

This is a little bit tricky, but possible. I can help with this and co-work if helpful. 

It looks like the standalone code generator from AJV supports multiple schema and output a single file with an export for each schema type, so that shouldn't need a build step (we can do `import * from './ajv-output'` at runtime to get all the validation functions, and can iterate over their keys).

For protobuf, it looks like we will need to generate our own index file that will be something like:

```
export { Observation } from './observation'
export { Preset } from './preset'
// etc.
```

In order to be able to correctly type the encode and decode functions, we will also need to generate a file that generates a typescript type `type MapeoRecord = Observation | Preset | // etc`

I think the typing will become easier if you have a separate function for transforming the type from json schema to the type from protobuf, and vice-versa. This will probably need a generated `.d.ts` file too, defining overloads something like (don't think this exact code will work):

```ts
import { Observation as ProtobufObservation } from './types/proto/observation'
import { Observation as JsonSchemaObservation } from './types/schema/observation'

function transformToProto(obj: JsonSchemaObservation): ProtobufObservation
function transformToProto(obj: JsonSchemaPreset): ProtobufPreset

export transformToProto

**Comment by tomasciccola** [comment]
Since dynamic import won't work at runtime, how can I serialize the functions generated by the protobufs? (encode, decode,etc.) when doing the code generation? 
The only think I can think is to do Function.toString() for each member of each protobuf schema. Does it makes sense or is Function.toString() cursed?? You can see an example at `scripts/genSchemas.js`

**Comment by gmaclennan** [comment]
> Since dynamic import won't work at runtime, how can I serialize the functions generated by the protobufs? (encode, decode,etc.) when doing the code generation?
> The only think I can think is to do Function.toString() for each member of each protobuf schema. Does it makes sense or is Function.toString() cursed?? You can see an example at `scripts/genSchemas.js`

I feel like it is kind of cursed, but I think there are valid uses for it. However, in this case it will not work because the functions reference helper functions that are in the global scope of the generated files (e.g. `function base64FromBytes()` in the file `types/proto/observation.js`), and those functions are not included when you do Function.toString().

I think you can do this by generating an index.js via code generation. It looks like the way the protobuf files are generated, each exports a named export with the (camel case?) name of the schema. So a generated index.js should look like:

```js
export { Observation } from "./observation.js"
export { Preset } from "./preset.js"
```

I don't think you will need to generate a `index.d.ts` for the typescript types to be picked up, because typescript will infer them from `index.js` and the `observation.d.ts` file.

You can then consume this via:

```js
import * as PbfSchemas from "./types/proto/index.js"

console.log(Object.keys(PbfSchemas))
// ['Observation', 'Preset']

/** @type {PbfSchemas.Observation */
const o = {
  id: // etc...
}
```

To generate this you can do something like:

```js
const lines = []
const files = await glob('./types/proto/*.js')
for (const filename of files) {
  const moduleExports = await import(filename)
  const exportNames = Object.keys(moduleExports)
  const generatedLine = `import { ${exportNames.join(', ')} from '${filename}'`
  lines.push(generatedLine)
}
const generatedCode = lines.join('\n') + '\n'
```

tomasciccola:
Yeah, a while after I posted I realize there was no need for what I said; it was sleep deprivation talking for me :|

tomasciccola:
Well, I've move some foward with this, mostly code generation for:
* having all the jsonSchema `validate` functions at runtime
* having a map from `doctype` to `{schemaVersion, dataTypeId}` mostly for usefulness
* wraping all jsonSchema types in a union type
* having an index.js that exports all the protobuf types from one place

Also some work on getting the encoding/decoding of the blockPrefix right. 
Regarding this, I'm using compact-encoding and there are some type errors: it seems that `cenc.encode` is typed to return null. Same with `state.buffer`: is initialized as null so ts complains when assigning a buffer to it. I don't know what to make of it...

TODO for this PR:
* having typed fns from JsonSchema -> ProtoBufSchema and viceversa 
* Having all schemas extend common.json schema. This will also solve some type errors, since fields like `type` (a string that indicates recordType) is only defined on common.json and not the rest (in fact, some other schemas have the same field `type` but use for smth else, so we should revise that).
* Finishing encode and decode fns
* getting ts to stop complaining
* Fix and add tests

tomasciccola:
Wouldn't it make more sense to avoid overloads and instead receive and return union types?
So instead of:
> 
> ```ts
> import { Observation as ProtobufObservation } from './types/proto/observation'
> import { Observation as JsonSchemaObservation } from './types/schema/observation'
> 
> function transformToProto(obj: JsonSchemaObservation): ProtobufObservation
> function transformToProto(obj: JsonSchemaPreset): ProtobufPreset
> 
> export transformToProto
> ```
We can do:
```ts
import { MapeoRecord as JsonSchemaRecord } from './types/schema/index'
import { MapeoRecord as ProtoSchemaRecord } from './types/proto/index'

function jsonSchemaToProto(obj:JsonSchemaRecord): ProtobufSchemaRecord
function protoToJsonSchema(obj:ProtobufSchemaRecord): JsonSchemaRecord
```

gmaclennan:
> Wouldn't it make more sense to avoid overloads and instead receive and return union types?

The behaviour will be slightly different. With overloads if you call `transformToProto(observation)` then the returned type will be `ProtobufObservation`, so it will be accepted by the protobuf encode function. With the second option the returned type will be `ProtobufSchemaRecord` so you would need to narrow it to `ProtobufObservation` before passing it to the protobuf encode function.

tomasciccola:
It's becoming increasingly difficult to debug some type errors that I've been having. Now they are related to slightly different expectations between types for jsonSchemas and types for protobufDocs (i.e. some fields in jsonSchema are `string | undefined` whereas in protobufDocs `string | null`). 
In order to avoid getting myself lost I'm going to nuke the jsonSchema folder on this branch and start from the common.json schema and observation.json schema (extending from common.json using $anyof - which I don't know if its the best approach but for now maybe it'll do). I'm going to do the same with protobuf (defining a Common protobuf and an Observation extending from Common - still need to look the best approach for this...). 
After that I'll probably add schemas and protobuf needed from authstore (I think @sethvincent you layed them out somewhere).
wdyt? I haven changed jsonschemas much so we can always re-grab them from the master branch

tomasciccola:
I have most of the stuff working (I think). Some jsonschemas don't have fields like schemaVersion or type, which are key to selecting the correct schema to validate. To solve for that, I've basically created a base type:
```ts
interface base {
 type: string;
 schemaVersion: number;
 [key:string]: any;
}
```
and then create a mix between an Union type (of all schemas) and an intersection with base (you can see this on `scripts/generate.js:77`).
That way, a user can still pass a type and schemaVersion despite not existing on schemas like preset. Does this make sense? @gmaclennan @sethvincent

tomasciccola:
The only odd one out of the schemas is schemas/field.json which doesn't have a schemaVersion field. Is this schema actually used on current/older mapeo versions? It would be nice to know so that I contemplate that case without a version on the schema (I could do something like grabbing the version from the path and 'making-up' that its version is v1 so that I don't break backwards compatibility...)

sethvincent:
@tomasciccola working on migration I noticed a few small things.

### first build
I got this type error:

types/proto/index.d.ts:2:30 - error TS1110: Type expected.

```ts
export type ProtobufSchemas =
```

It seems like this empty string after the equal sign problem happens on first `npm run build` but after running a second time this file is generated successfully with the full type definition.

Similarly it seems that creating/destroying the `dist` directory and maybe other generated directories needs to be part of the npm scripts.

### toString

for some reason typescript expects no argument for toString() so i had to add this:

```js
    // @ts-ignore
    id: common?.id.toString('hex') || '', 
```

### convert id to string

in `protoToJsonSchema` it looks like there's a type error where this uint8 array needs to be converted to a string:

```js
const protoToJsonSchema = (protobufObj, { schemaVersion, type, version }) => {
  const common = protobufObj.common
  delete protobufObj.common
  return {
    ...protobufObj,
    ...common,
    schemaVersion,
    type,
    version,
    // @ts-ignore
    id: common.id.toString('hex')
  }
}
```

tomasciccola:
Thanks for the comments!

I fixed the script/generate.js to take into acount the absence of dist/ and to use .ts files in types/proto (since tsc is run after)

> ### toString
> for some reason typescript expects no argument for toString() so i had to add this:

Its weird that I don't get that type error!
and the same happens to the `protoToJsonSchema` fn. I don't get type errors with the id as string.
But I'll double check if there's anything weird there...

I realize that we should also version the protobufs and generate corresponding ones for older versions of jsonSchemas that for example don't extend `common.json` schema. I realize this when trying to add other protobufs (like `preset.json`)

tomasciccola:
So, leaving behind the comments that @sethvincent made regarding type error when doing `Buffer.toString('hex')` (which doesn't show as a type error on my computer :|  ...), I think this is maybe somewhat close to being merged?
I think there's significant work to do regarding schemas. 
As I mentioned on my past comment, maybe we should also version protobuf schemas, and there's some other stuff like adding some fields to protobuf schemas to match some record types already present in jsonschemas.
Maybe it should make sense to add some schemas used by auth store (and contemplate them on the tests), but besides that I think it would make sense to close this PR and continue working on this on a different PR

package.json
```diff
@@ -2,13 +2,19 @@
...
+    "example": "node examples/schema_test.js",
+    "protobuf": "buf generate ./proto",
+    "jsonschema": "scripts/./jsonts",
+    "tsc": "tsc",
+    "tsc-index": "tsc index.js -d --emitDeclarationOnly --allowJs --allowSyntheticDefaultImports --module es2022 --target es2019 --moduleResolution node",
+    "generate": "node scripts/generate.js",
+    "doc": "typedoc --plugin typedoc-plugin-markdown",
+    "clean": "rm -rf types dist docs",
+    "build": "npm-run-all protobuf jsonschema generate tsc tsc-index doc",
+    "test": "tape 'test/**/*.js'"
```
sethvincent, line 17:
```suggestion
    "prepare": "npm run build",
    "test": "tape 'test/**/*.js'"
```

This will make it so the schemas get built on `npm install`. I think in the future we could do this step (and `npm publish`) in a github actions job, but this works for now.

sethvincent:
This is looking good! Going ahead with the merge. We can create issues and follow-up prs for anything else that comes up.

PR #26, opened 2023-01-16
feat!: update export names to improve type definition exports

achou11:
1. Adds the `types` field in `package.json`

2. Updates the build script to also export the TS interfaces for each entity in `./types/`

3. Updates the exported definition objects from the module to have a `Schema` suffix, e.g. `Common` is now `CommonSchema`. **This is a breaking change** although we can potentially avoid it by renaming the exported interfaces from the `./types/` directory instead, which I think is a little more difficult.

Note that this is pointing to `typescript` branch, not master

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -10,6 +10,18 @@ export { validateFilter };
 export { validateObservation };
 export { validatePreset };
 
+import CommonSchemaDef from './schema/common.json';
+import FieldSchemaDef from './schema/field.json';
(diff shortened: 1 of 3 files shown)


achou11:
Hm just realize that the schema objects exported from `index.js` are actually jsonschema objects, and are thus not with the interface definitions in `./types`. will have to be a bit smarter about how we export in this case

gmaclennan:
Resolved in master

PR #29, opened 2023-02-09
allow lowercase type name in documents

sethvincent:


tomasciccola:
I know this is already merged, but the fn you add already exists on scripts/generate.js:21 as `capitalize`. I'll probably move it to a utils.js file on a future PR and import it from there

sethvincent:
oh! sounds good

Issue #39, opened 2023-03-09
Type imports/exports


sethvincent:
Importing a type is currently kinda lengthy.

In our jsdoc types long paths get messy looking and are kinda hard to guess:

Currently it's something like:

```js
/**
* @property {import('mapeo-schema/types/schema/observation/v5').Observation} Observation
*/
```

Ideally we'd get it down to something like:

```js
/**
* @property {import('mapeo-schema/types').Observation} Observation
*/
```

Importing types in this way would likely require generating an index.js/index.d.ts in the `types` directory and using the [`exports` field in package.json](https://nodejs.org/api/packages.html#package-entry-points) to map `/types` in the import string to the `./types/index.js` file. (I'm pretty sure this is how the above import examples could be achieved though haven't double checked)

tomasciccola:
In that ideal import its not clear if we're referring to JSONSchema or ProtobufSchema, I would guess JSONSchema since that is what we want to expose to the user, but do you think that's the case??

tomasciccola:
I mean, `types/index.d.ts' could be smth like
```ts
export { ProtobufSchemas } from './proto'
import { JSONSchemas } from './schema'
export default JSONSchemas
```

but importing from JSdoc would be
```js
/**
* @property {import('mapeo-schema/types').default.Observation} Observation
*/
```

or
```js
/**
* @property {import('mapeo-schema/types').JSONSchema.Observation} Observation
*/
```

sethvincent:
closed by #43

Issue #38, opened 2023-03-09
Type naming


sethvincent:
A couple of things to decide related to type naming:

- should we include the schema version in the type name?
- should we include the encoding in the type name? json vs protobuf

Some notes I had on this from the #32 pull request:

It might be ideal to have a naming convention like:

- `ObservationJson`
- `ObservationProtobuf`

Related: it looks like the protobuf comments export names like `CoreOwnership_1` rather than `CoreOwnership`.

I have mixed feelings on whether the public types should include the version in the name. The main place that might matter is in migration code where it's needed to be explicit about versions and use multiple versions in the same code.

Maybe we could import latest version as just the name:

```js
/**
* @property {import('mapeo-schema/types').ObservationJson} Observation
*/

/**
* @property {import('mapeo-schema/types').ObservationProtobuf} Observation
*/
```

As well as being more specific about the version needed:

```js
/**
* @property {import('mapeo-schema/types').ObservationJson4} Observation
*/

/**
* @property {import('mapeo-schema/types').ObservationJson5} Observation
*/
```

A type without a version could be the latest version.

I can imagine there's a good argument for always including the version number and not having the convenience of importing the latest version by just the schema name.

tomasciccola:
protobuf names should be unique for a project, so that's the reason they have the version on the name. I'll probably rename the export so that the last version is the default one with the possibility of importing a specific version

sethvincent:
closed by #43

PR #43, opened 2023-04-05
better exporting of types

tomasciccola:
This PR addresses #39 and #38

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
  * @module mapeo-schema
  */
 import * as cenc from 'compact-encoding'
-import * as JSONSchemas from './dist/schemas.js'
+import * as Schemas from './dist/schemas.js'
 import * as ProtobufSchemas from './types/proto/index.js'
 import schemasPrefix from './schemasPrefix.js'
 import { inheritsFromCommon, formatSchemaKey } from './utils.js'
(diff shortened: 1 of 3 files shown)


tomasciccola:
I basically now generate a `types/index.d.ts` with the following contents:
```js
export { ProtobufSchema } from './proto'
export { JSONSchema } from './schema'
```
both refer to a mixed type that can the be imported like:
```js
/**
* @property {import('./types').JSONSchema} obj
*/
```

However one cannot access individual schemas from there like so:
```js
/**
THIS WON'T WORK
* @property {import('./types').JSONSchema.Observation} observation
*/
```
Since i.e Observation is not actually a field inside JSONSchema. 
To import individual types one would do smth like
```js
/**
THIS WILL WORK
* @property {import('./types/schema').Observation} observation
*/
```

this works since each `index.d.ts` corresponding to JSONSchemas and Protobufs also export individual types referring every version of a schema or protobuf, where the identifier without a specific version corresponds to the last version (so `Observation` actually refers to `Observation_5`).

tomasciccola:
An alternative to referring to individual types would be exporting every JSONSchema and Proto types from `types/index.d.ts`, where we append an identifier like so
```js
export { Observation_5 as ObservationProto } from './proto/observation/v5'
export { Observation_4 as ObservationProto_4 } from './proto/observation/v4'
``` 
this, I think, would allow to do smth like
```js
/**
* @param {import ('./types').ObservationProto} observation
*/
```
instead of
```js
/**
* @param {import ('./types/proto').Observation} observation
*/
```

I don't have any preferences, so open to suggestions...

sethvincent:
> ```js
> /**
> THIS WILL WORK
> * @property {import('./types/schema').Observation} observation
> */
> ```
> 
> this works since each `index.d.ts` corresponding to JSONSchemas and Protobufs also export individual types referring every version of a schema or protobuf, where the identifier without a specific version corresponds to the last version (so `Observation` actually refers to `Observation_5`).

Cool, that works for me. I'll do some quick experiments using/integrating this.

PR #44, opened 2023-05-30
Update mapeo-schema for integration with mapeo-core-next

sethvincent:
This pull request is a dependency for another PR in mapeo-core-next that integrates @mapeo/schema: https://github.com/digidem/mapeo-core-next/pull/111


proto/role/v1.proto
tomasciccola, line 1: does type refer to the type of doc? aren't we encoding this on the header?

tomasciccola: aahh, ok. For me this PR is ready to merge <3

gmaclennan:
Should be resolved by #93

PR #49, opened 2023-06-27
chore: add npm release action

gmaclennan:
FIx #45 by adding a github action to automate releases and publishing to npm


gmaclennan:
This needs to be in the default branch in order to test the workflow_dispatch trigger to initiate a release. I'm going to go ahead and merge without review since this does not change anything else in the code that could break things.

PR #56, opened 2023-06-27
chore: update deps

gmaclennan:
Updates "low risk" deps that I don't think will affect code behaviour. If tests pass then I think this is ok to merge. There are also updates to typescript, @sinclair/typbox and typedoc which I think are higher risk and we can look at updating separately.


gmaclennan:
replaced by #90

Issue #46, opened 2023-06-22
Schemas for config store


tomasciccola:
Linked to [this](https://github.com/digidem/mapeo-schema/issues/25) and [this](https://github.com/digidem/mapeo-core-next/issues/78) issue.
To start off, I'm using [this](https://github.com/digidem/mapeo-core-next/pull/92/files/47f3f2afc4d3bd6a3af2b010bd30bda9cced5848#diff-e2d245dad95ae2fdda6c5c8984eaa16226d9dc308e39ec65c695d3d49237c367) file in the ConfigStore PR as a reference.

PR #65, opened 2023-07-12
Project record type

tomasciccola:
This PR adds the `Project` record type. I only added the `name` field, but later we should think about other fields that could be useful. 
This is linked to [this issue](https://github.com/digidem/mapeo-schema/issues/46), it can close it when merge it (and we should create another issue to track work on the `Project` record type) or we can keep it open


achou11 (approved):
lgtm!

Issue #42, opened 2023-03-20
Field v1: replace `any` in `key` with `oneof`


tomasciccola:
Since the `key` field in the schema `Field` v1 is a string OR and array I initially encoded as an `any` type in the protobuf. 
But protobufs has `oneof` to union multiple fields. 
I've already have most of it done, but some errors still persist. Just opening this issue to come back to it later

tomasciccola:
@sethvincent @gmaclennan do y'all think this is worth working on right now?

PR #66, opened 2023-07-12
feat: field v2 using key as always an array of strings

tomasciccola:
This PR creates a `field/v2` protobuf and json schema to set the `key` field to be a always an `Array` of  type `string`. 
I took the liberty to modify the `field/v1` protobuf to use `repeated string key` instead of `anyof` (since there's no backwards compatibility to keep here)
This protobuf is exactly the same as `field/v2`, but I keep it to mantain parity with the json schema versioning. 
To take into account the `anyof` of `field/v1` jsonschema I basically check if the key field is a `string` and the record is `Field_1` and wrap the `string` in an `Array`.
Since the return value is always an `Array` of `string` then JSONSchema validation will pass.
This should close [this issue](https://github.com/digidem/mapeo-schema/issues/42)

### DIFF
--- a/examples/schema_test.js
+++ b/examples/schema_test.js
@@ -69,6 +69,7 @@ const obj = {
   id: randomBytes(32).toString('hex'),
   schemaType: 'Field',
   schemaVersion: 1,
+  created_at: new Date().toJSON(),
   key: ['hi'],
   type: 'text',
 }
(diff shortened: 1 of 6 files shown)


gmaclennan:
Decided to not support nested tags, and limit `tagKey` to be a string.

Issue #70, opened 2023-07-24
Finalize schema definitions for data & config core records


gmaclennan:
## Description

Finalize the Protobuf and JSONSchema definitions for:

- field
- preset
- project
- observation
- common

## Not included

- coreOwnership
- device
- role

These are pending review of capability code & tests

gmaclennan:
Done in https://github.com/digidem/mapeo-schema/tree/schemaRevision - needs separated and merged

Issue #64, opened 2023-07-12
Finalize message type GUIDs


gmaclennan:

tomasciccola:
- [ ] Generate random GUIDs for every record type. This is stored on `schemasPrefix.js` as 
```js
{[schemaType]:{ dataTypeId: string, schemaVersions: [number]}}
```
(Can we avoid future modifications of this id's? One can be a runtime check where `schemasPrefix` is an immutable Object (Object.freeze?). But also we can freeze the file in [git](https://stackoverflow.com/questions/23528104/how-to-freeze-a-file-in-a-repository) ?)
- [ ] Revise and ***document*** criteria for `dataTypeId` size taking into account probabilities of hash collisions in terms of number of posible `dataTypeId`s generated
- [ ] Document format of header of saved records

Issue #36, opened 2023-02-27
Schema questions


gmaclennan:
### Protobuf

- Store `links` as `Array<{ coreKey: Buffer, index: number }>`?
- Store `created_at` and `timestamp` as Timestamp or number vs. string?
- Rename `timestamp` to `edited_at`?
- Camelcase props? e.g. `createdAt`?
- `deviceId` as bytes
- Define type of observation refs
- Define type of observation attachments (since we can store metadata in the blobstore now, how much info here?)

### JSON Schema

- `created_at` and `timestamp` as type `number` or `Date`?
- Do we need to return the `schemaVersion` field? Maybe the contract with the client is that the latest schema version is always returned, and is typed.
- Also store `version` for `refs`?
- Restrictions on `tags`? E.g. JSON? Nested?

### Other types

- Protobuf for `node`, `way`, `relation`, `preset`, `field`?
- `device` and `project` record types needed?
- Anything else? `ownership`, `capability`?
- Mapping older schema versions - can do most at migration to avoid runtime migration?

sethvincent:
> Store links as Array<{ coreKey: Buffer, index: number }>

Not needing to split on `@` sounds awesome to me.

> Define type of observation attachments

It seems like this could now just be whatever we need to do a get. Or whatever we need to fetch the metadata.

> Restrictions on tags? E.g. JSON? Nested?

Is there config with json value tags in use now? Nested json is kinda scary.

---

Yeah timestamp should be renamed to either updated_at or edited_at. updated_at is more common.

Using number type for any date fields makes sense to me.

Lately I keep daydreaming about how nice it would be if everything in js was snake_case. It's so readable! But, for consistency I'd lean toward making things camelCase.

gmaclennan:
To bring in some ideas from the Mapeo Core API discussion:

- Rename `id` to `docId` and `version` to `versionId`
- Use `updatedAt` instead of `timestamp` (agreed with Seth about snake_case, but let's stick with camel for consistency)

> Store links as Array<{ coreKey: Buffer, index: number }>

[...]

PR #72, opened 2023-07-24
fix!: revise schema for data & config store records [1/6]

gmaclennan:
Extracted from #67 to just include changes to JSONSchema and Protobuf, not changes to the encode/decode code and generation scripts. Only focusses on record types from the data and config stores:

- field
- preset
- project
- common
- observation

## Tasks

- [x] Common Protobuf name and field changes
  - `created_at` - > `createdAt`
  - `deviceId` - remove (derive at runtime)
  - `links` → repeated bytes
  - `timestamp` → `updatedAt` (shouldn’t be an optional)
  - `userId` - remove for now
 - [x] Common JSONSchema name and field changes
    - [x]  `id` - “hex-encoded 32-byte buffer”
    - [x] `version` - “coreId (hex-encoded 32-byte buffer) and core sequence number, separated by `/`”
    - [x] `created_at` - > `createdAt`
    - [x] `timestamp` → `updatedAt` (shouldn’t be an optional)
    - [x]  `deviceId` and `userId` - remove from jsonschema
    - [x] `schemaVersion` - remove? Assume `encode` only accepts most recent schema.
    - [x] `schemaType` -> `schemaName` - original `schemaType` was confusing for naming variables.
    - [x]  `required` - add `UpdatedAt` (On created it should be the same as createdAt, ). and change `created_at`. add `links`.
  
 - [x] Observation Protobuf name and field changes
     - `refs` → repeated message with { id: bytes }
    - `attachments` → repeated message: look at [[BlobId](https://github.com/digidem/mapeo-core-next/blob/main/lib/types.d.ts#L13-L22)](https://github.com/digidem/mapeo-core-next/blob/main/lib/types.d.ts#L13-L22) but without the `variant` - { driveId: string, name: string, type: ‘photo’ | ‘video’ | ‘audio’ }
    - `tags` → not optional.
    - `manual_location` → 🐫
    - Put everything from JSONSchema in the proto definition (see v4.proto)
 - [x] Observation JSONSchema name and field changes
     - tags to be stricter
  - [x] Find a way to avoid type errors when `JSONSchema->Proto` or avoid copying the object
  - [x] Revise Field
  - [x] Delete Filter
  - [x] Revise Preset


gmaclennan:
Fixes
- #36 
- #64 
- #42 
- #70

tomasciccola (approved):

tomasciccola:
Checks are failing because it seems that the buf version we use in the CI doesn't support `[(required) = true]`

PR #77, opened 2023-07-27
fix: Change prop names and type names [5/6]

gmaclennan:
Following discussion with @achou11:

- Rename `id` -> `docId`
- Rename `version` -> `versionId`
- Export types for "values", e.g. without common props. These are the types for update/create actions.
- Name "value" types as `${schemaName}Value`
- Rename union of all Mapeo types as `MapeoDoc`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mapeo/schema",
-  "version": "3.0.0-next.3",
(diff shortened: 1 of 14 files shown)


gmaclennan:
Actual squash-merge commit 73bda106d89d66f8930f100c3019e7b0675d4736

PR #78, opened 2023-07-27
feat: opaque types [6/6]

gmaclennan:
This ensures that the client does not try to generate these themselves,
and avoids errors from assigning another string to these values. They
should always come from database reads, and their use should be scoped
to documents of the same schema type. The opaque types ensure that.


gmaclennan:
Re-visiting this, I think we should skip opaque types for now, because they are going to create headaches in the mapeo-core codebase. If we think they will be useful in the frontend client we can cast them there.

PR #80, opened 2023-07-27
chore: fix up TS generation and build for npm package [4/6]

gmaclennan:
Needed to jump through a few hoops to make Typescript happy:

- Uses the `rootDirs` option so that the `src` and `generated` folders appear as merged to Typescript, so that you can import generated code from code in `src` as if it was in the same folder.
- Needed to output compiled Typescript into an `intermediate` folder, and then merge with a separate build step (limitation of Typescript)

This also specifies to only include the `dist` folder in the npm package, and changes the script to build everything from `prepublishOnly` to `prepack`.


gmaclennan:
Actual squash-merge commit 146a6fbbc56b6f70e5a3aca2d1fe0ba8eee022bf

PR #93, opened 2023-08-02
feat: add missing fields to authStore schemas/protos

tomasciccola:
This PR adds missing fields to authStore Schemas and protobufs. 
Things that can be revised:
* some fields name *Id that are typed as `string` on the protobuf could be of type `bytes`.
* `capabilityType` is a `string` but can be an `enum`. Looking at [this code](https://github.com/digidem/mapeo-core-next/blob/main/lib/authstore/authtypes.js) I don't know this should be the name of the role (`projectCreator | coordinator | member | non-member`) or the capabilities (`read | write | manage:devices`).

This should close PR #44

### DIFF
--- a/proto/coreOwnership/v1.proto
+++ b/proto/coreOwnership/v1.proto
@@ -20,4 +20,6 @@ message CoreOwnership_1 {
   string signature = 9;
   int32 authorIndex = 10;
   int32 deviceIndex = 11;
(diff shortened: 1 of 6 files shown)


gmaclennan (approved):
Looks good to me. Ok to leave possible enums as strings for now, and we can review later.

PR #95, opened 2023-08-02
fix: change schemaName from enum to const

gmaclennan:
We had been using `enum` to define the `schemaName` property, however this was proving challenging to work with in the Mapeo Core code - it's easier if we use the `const` keyword, which is more appropriate for this (although they essentially do the same thing).


tomasciccola (approved):
yeah, this change makes a lot of sense!

PR #96, opened 2023-08-02
Conversion functions for authStore types

tomasciccola:
Things to consider:
* what fields are required? 
  * Role: every field
  * Device: every field
  * coreOwnership: every field
* Device: actions are strings. When we know possible actions, it should be an enum.
* action name delimiters:
  protobuf doesn't allow colons (':') on identifiers, so to avoid type problems I basically replaced that delimiter with underscore ('_'). So in schemas `role:set` becomes `role_set`. Another separator could be chosen, or a convertion function (convertAction) could be written; I've tried but failed with that

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -956,7 +956,7 @@
     "concat-map": {
       "version": "0.0.1",
       "resolved": "https://registry.npmjs.org/concat-map/-/concat-map-0.0.1.tgz",
-      "integrity": "sha1-2Klr13/Wjfd5OnMDajug1UBdR3s=",
+      "integrity": "sha512-/Srv4dswyQNBfohGpz9o6Yb3Gz3SrUDqBH5rTuhGR7ahtlbYKnVxw2bCFMRljaA7EXHaXZ8wsHdodFvbkhKmqg==",
       "dev": true
     },
     "convert-source-map": {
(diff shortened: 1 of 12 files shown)


gmaclennan (changes requested):
Looks good to me, I think the enums with `_` are fine.

I think the JSONSchema for Role should include `UNRECOGNIZED` as an enum option, rather than trying to use a default, and the client should handle that accordingly. This is for future compatibility so that a new action does not break existing clients.

Needs changes from master merged in, for the const: schemaName change.

tomasciccola:
> Looks good to me, I think the enums with `_` are fine.
> 
> I think the JSONSchema for Role should include `UNRECOGNIZED` as an enum option, rather than trying to use a default, and the client should handle that accordingly. This is for future compatibility so that a new action does not break existing clients.
> 
> Needs changes from master merged in, for the const: schemaName change.

Ok, I see that, for example, the `Field` schema doesn't have an `UNRECOGNIZED` option on the `type` field (only `text|number|selectOne|selectMultiple`) in what sense that case is different? mostly because we're certain we are not going to add additional options in the future??

gmaclennan:
> Ok, I see that, for example, the `Field` schema doesn't have an `UNRECOGNIZED` option on the `type` field (only `text|number|selectOne|selectMultiple`) in what sense that case is different? mostly because we're certain we are not going to add additional options in the future??

Ah good point. I initially had initially thought that a `text` field is a good fallback for anything, but because the actual values stored in the tags could be different, I think it's better for the client to decide how to deal with this, by which I mean that `UNRECOGNIZED` is a good option here - if the client sees that, it can look at any existing value and choose the best way to display it.

It's a different case with the `appearance` option, where `multiline` is a good fallback for any future options.

gmaclennan (approved):

PR #99, opened 2023-08-02
feat: add defaultPresets to Project doc

tomasciccola:
Added `defaultPresets` as an object with `point (string[])` and `area (string[])`.

Since project has two versions for testing foward compatibility I added the field to both of them, but it maybe a good choice to only add it to the newest (v2) so to test foward compatibility on conversion functions (maybe outside of the scope of this tickets). I don't know if having a check for `schemaVersion` on the `decode-conversions.ts/convertProject` function would suffice (it seems like it...)

### DIFF
--- a/proto/project/v1.proto
+++ b/proto/project/v1.proto
@@ -12,19 +12,12 @@ message Project_1 {
 
   Common_1 common = 1;
 
+  message DefaultPresets {
+    repeated string point = 1;
+    repeated string area = 2;
+  };
+
+  DefaultPresets defaultPresets = 2;
+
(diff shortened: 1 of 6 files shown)


gmaclennan (changes requested):
Looks good, just a couple of suggestions:

- I think keep `vertex`, `line`, `relation` as options - we do use them sometimes with presets for iDEditor (Mapeo territory view)
- I think on the protobuf make these repeated bytes, since these are actually preset `docId`s - for no other reason than consistency.

gmaclennan (approved):

PR #100, opened 2023-08-03
feat: export fns for versionId, seq->index, coreId->coreKey

gmaclennan:
Exports functions for parsing a string `versionId` to get the core key
and index, and for generating a string versionId. Also renames the
parameters: The naming convention of `seq` goes back many years to when
we used Hyperlog (precursor to Hypercore). `index` is the naming
currently used in Hypercore docs and I think it makes the most sense -
it refers to the index of the block within the hypercore. We also have a
convention in Mapeo Core of using `key` to refer to a Buffer, and `id`
to refer to a string. `docId` is the exception when named on the
protobuf, because it doesn't make sense to rename it when decoding, and
because `docId` as a string is never exposed outside this library.

### DIFF
--- a/examples/schema_test.js
+++ b/examples/schema_test.js
@@ -20,7 +20,7 @@ async function test(key) {
     const index = 0
     const data = await core.get(index)
     console.log(`trying ${k}`)
-    const decodedData = decode(data, { coreId: core.key, seq: index })
+    const decodedData = decode(data, { coreKey: core.key, index })
     console.log('data', decodedData)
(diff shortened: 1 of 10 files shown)


tomasciccola:
> We also have a
> convention in Mapeo Core of using `key` to refer to a Buffer, and `id`
> to refer to a string. `docId` is the exception when named on the
> protobuf, because it doesn't make sense to rename it when decoding, and
> because `docId` as a string is never exposed outside this library.

Does this applies to every field with `Id` as a suffix?

tomasciccola (approved):

PR #106, opened 2023-08-07
Fix Tests

### LINKED ISSUE #79: Fix tests
Tests need updated good/bad docs to match the schema updates

tomasciccola:
This should close #79.
Things I've noticed when writing different `MapeoDoc`s  for testing:
* passing `docId` fails JSONSchema validation (JSONSchema thinks is an additional - not allowed - field)
* We are not handling JSONSchema validation errors, we could throw them or return them (which would change the `validation` fn signature) 
 ```js
  const validationFn = validations[schemaName]
  if(validationFn(obj)) return true
  throw new Error(validationFn.errors.map(err => JSON.stringify(err)).join('\n'),null,4)
```

I'm missing more tests:
* check equality of decoded object with passed object
* write a bunch of docs with different missing properties and invalid types for fields
* test decoding a buffer with wrong random header
* test decoding a buffer with header with wrong schemaType and/or schemaVersion
* test JSONSchema validation
* smth missing?

Also, not important nor a priority, but we're using `tape` for testing. We could change to using `brittle` since they are pretty similar. I don't have a rationale for that change, but mostly for the sake of uniformity in tools...

### CODE REVIEW

**Comment by gmaclennan** [comment]
> * passing `docId` fails JSONSchema validation (JSONSchema thinks is an additional - not allowed - field)

Currently the validation function validates the "valueSchema", e.g. the schema without the common properties. This is because the main place where we validate data is on user (API) input, and the API never passes the common properties - those are set on the server. I don't _think_ there is a need to validate objects that include the common properties?

> * We are not handling JSONSchema validation errors, we could throw them or return them (which would change the `validation` fn signature)

I think it would be good to do the same as Ajv here:

```js
export function validate(schemaName, obj) {
    const isValid = validations[schemaName](obj);
    validate.errors = validations[schemaName].errors
    return isValid
}
```

> 
> ```js
>  const validationFn = validations[schemaName]
>  if(validationFn(obj)) return true
>  throw new Error(validationFn.errors.map(err => JSON.stringify(err)).join('\n'),null,4)
> ```
> 
> I'm missing more tests:
> 
> * check equality of decoded object with passed object
> * write a bunch of docs with different missing properties and invalid types for fields
> * test decoding a buffer with wrong random header
> * test decoding a buffer with header with wrong schemaType and/or schemaVersion
> * test JSONSchema validation

I don't think this needs to be thorough - the Ajv tests should catch that. We just need to check that a correct document of each schema type validates correctly, to make sure we're exporting the right thing.

> * smth missing?
> 
> Also, not important nor a priority, but we're using `tape` for testing. We could change to using `brittle` since they are pretty similar. I don't have a rationale for that change, but mostly for the sake of uniformity in tools...

We use `tape` in almost everything else, I'm ok with sticking to that. `brittle` has been ok to use in mapeo-core-next, but I don't think it adds anything that we need over tape.

gmaclennan (commented):
Progress looks good. A suggestion about how to write this to avoid wrapping test functions:

```js
test('bad docs throw', t => {
    t.plan(fixtures.badDocs.length)
    for (const { doc, text } of fixtures.badDocs) {
        t.throws(() => encode(doc), text)
    }
})
```

```js
test('good docs encode and decode correctly', t => {
    t.plan(fixtures.goodDocs.length * 2)
    for (const { doc, text } of badDocs) {
        let buf
        t.doesNotThrow(() => {
            buf = encode(doc)
        })
        const decoded = decode(buf, parseVersionId(doc.versionId))
        t.deepEqual(decoded, doc)
    }
})
```

FYI if you use pass an async function to `test()`, then you don't need to make a plan - it will consider the test finished when the function resolves. You can do that, or make the plan like I do above.

deepEqual might fail because of the difference between undefined properties and uninitialized properties, e.g.

```
const a = {
  optionalProp: undefined
}
const b = {}
```

Those won't be deep equal, but will be valid as far as our schema and the type are concerned. Probably need a different deep equal function for testing this.

test/fixture.js
```diff
@@ -0,0 +1,136 @@
...
+export const fixtures = {
+  onlyId: {
+    text: 'test encoding of record with missing fields',
+    plan: 1,
+    doc: {id: randomBytes(32).toString('hex')}
+  },
+  badDocName: {
+    text: 'test encoding of wrong record type',
+    plan: 1,
+    doc: {
```
gmaclennan, line 12: Everything else should be correct on this doc if we want to test the schemaName. E.g. needs a `docId`, `versionId` and `updatedAt`.

test/fixture.js
```diff
@@ -0,0 +1,136 @@
+import { randomBytes } from 'node:crypto'
+
+export const fixtures = {
+  onlyId: {
+    text: 'test encoding of record with missing fields',
+    plan: 1,
+    doc: {id: randomBytes(32).toString('hex')}
```
gmaclennan, line 7: maybe change to `docId`, since testing missing fields not invalid fields.

gmaclennan:
> deepEqual might fail because of the difference between undefined properties and uninitialized properties

I came across this same issue today when writing tests. I solved it the "dumb" way by [round-tripping the object to JSON](https://github.com/digidem/mapeo-core-next/blob/cf0b5350da7a7019062c666100c25e892ee24424/test-e2e/crud.js#L51) in order to remove undefined properties.

e.g. you can do:

```js
function stripUndef(obj) {
  return JSON.parse(JSON.stringify(obj))
}

t.deepEqual(stringUndef({ foo: undefined }), {}) // passes
```

test/fixture.js
tomasciccola: Yeah, that field was copy/pasted from older tests, I should change that.

tomasciccola:
> ```js
> function stripUndef(obj) {
>   return JSON.parse(JSON.stringify(obj))
> }
> 
> t.deepEqual({ foo: undefined }, {}) // fails because `foo` is missing from second parameter
> t.deepEqual(stringUndef({ foo: undefined }), {}) // passes
> ```

The issue I'm having with this is that, despite stripping `undefined` values, in JS `[] !== []` so `deepEqual` will fail if the object pass to encoding and the object we get back from decoding has a field with an empty array

tomasciccola:
Another issue I'm having is for decimals on floats. 
i.e.: when encoding `lat` in `observation`, if I encode a value like `-20.2` I would get back from decoding smth like `-20.2244782`

gmaclennan:
> The issue I'm having with this is that, despite stripping undefined values, a field which is an object where fields inside it are Arrays, comparison will fail.

This might actually be a bug. In general I've made array and object properties required to avoid this problem.

gmaclennan:
Decimal issue I've also came across. See https://github.com/digidem/mapeo-core-next/issues/150 - this suggests it might be a protobuf issue rather than SQLite. Maybe both. We could change how we encode floats - is there a more precise option? Or we could go for an option laid out in the issue (integer or string)

tomasciccola:
> > The issue I'm having with this is that, despite stripping undefined values, a field which is an object where fields inside it are Arrays, comparison will fail.
> 
> This might actually be a bug. In general I've made array and object properties required to avoid this problem.

I had a bug in `src/lib/encode-conversions.ts:31-33` which was tripping me off. Its fixed now...



> Decimal issue I've also came across. See [digidem/mapeo-core-next#150](https://github.com/digidem/mapeo-core-next/issues/150) - this suggests it might be a protobuf issue rather than SQLite. Maybe both. We could change how we encode floats - is there a more precise option? Or we could go for an option laid out in the issue (integer or string)

Its seems to be on the protobuf side since there's no sqlite here. Its weird that I haven't found an issue related to this on `ts-proto`

gmaclennan:
I think actually we should at least be using `double` first non-integers. I see `float` is only 32 bit. That's probably the source of the large inaccuracies. Let's make that change first and see how it changes things, then look at integer or string encoding if needed.

tomasciccola:
> I think actually we should at least be using `double` first non-integers. I see `float` is only 32 bit. That's probably the source of the large inaccuracies. Let's make that change first and see how it changes things, then look at integer or string encoding if needed.

Yeah, using doubles seems to fix the issue.

On the other hand, now I'm having another issue.
If I encode an object and avoid setting some values, i.e.:
```js
// here I'm not passing appearance, snakeCase, options and universal
    {
      docId: randomBytes(32).toString('hex'),
      versionId: `${randomBytes(32).toString('hex')}/0`,
      schemaName: 'field',
      createdAt: new Date().toJSON(),
      updatedAt: new Date().toJSON(),
      links: [],
      tagKey: "myTagKey",
      type: "text",
      label: "myLabel"
    },

```
protobuf won't throw an error, but also I would get back from decoding the same doc BUT with the added values (this are added on the conversion functions before encoding) which means that a comparison would fail (and so the test). This failing tests could be a good thing (hey! you're missing some fields), but in that case, shouldn't we avoid adding the missing fields with their default value and letting the encoding fail??

gmaclennan:
> protobuf won't throw an error, but also I would get back from decoding the same doc BUT with the added values

Yes, this is what we want and how it is coded - undefined fields get set to their defaults. The problem is that the test is using the input value to the encode() function as the expected value of the decode() function. This is not always the case.

I suggest that your fixture `goodDocs` includes `expected` where needed, and also a field describing what you are testing like with the `badDocs` e.g.

```js
goodDocs: [{
  text: 'Missing properties are set to defaults'
  doc: { ... }
  expected: { ... }
}, {
 ///
```

Then in the test:

```js
t.deepEqual(decodedDoc, entry.expected || entry.doc, entry.text)
```

Side-note, also consider using `for...of` rather than `array.forEach()`. Some [advantages listed here](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md) and it avoids the need to type the callback. `forEach` is fine too if you prefer that.

gmaclennan:
FYI at some point (not in these tests, but in Mapeo Core tests), this could cause an error:

```js
      createdAt: new Date().toJSON(),
      updatedAt: new Date().toJSON(),
```

Because `createdAt` and `updatedAt` should be the same for a newly created doc, but done like this I believe these _could_ end up different by 1ms. Not an issue in these tests, just to bear in mind if you end up testing something like this - it's a nasty kind of bug to figure out.

test/index.js
```diff
@@ -1,88 +1,128 @@
...
-})
+}
 
-test('test encoding of rightfully formated record', async (t) => {
-  const goodDocs = Object.keys(docs.good)
-  t.plan(goodDocs.length)
-  goodDocs.forEach((k) => {
-    const doc = docs.good[k]
-    t.doesNotThrow(() => {
+test('testing rightfully encoding of doc',async(t) => {
```
gmaclennan, line 29: This test is testing the same as the test below, we can remove it.

test/index.js
```diff
@@ -1,88 +1,128 @@
...
-      coreKey: randomBytes(32),
-      index: 0,
+test('testing encoding of doc, then decoding and comparing the two objects', async(t) => {
+  fixtures.goodDocs.forEach(
+    /** @param {import('../dist/types').MapeoDoc} doc */
+    (doc) => {
+      t.doesNotThrow(() =>  {
+        const buf = encode(doc)
+        const decodedDoc = stripUndef(decode(buf, parseVersionId(doc.versionId)))
+        t.deepEqual(doc,decodedDoc)
```
gmaclennan, line 46:
```suggestion
        t.deepEqual(decodedDoc, doc)
```

Function signature is `t.deepEqual(actual, expected)`.

gmaclennan: Move this outside the `doesNotThrow` callback, to be clear about what we are testing.

gmaclennan (changes requested):
For data type, we should test at least three "good docs":

1. Minimal doc with only required fields
2. Complete doc with all fields
3. Doc with extra fields (extra fields should not be encoded, so we will need a manually defined "expected" doc after the round trip).

For docs with default fields, we should check that not setting these and then decoding results in default values being set.

tomasciccola:
Some significant changes on the last commit:
* in `observation` proto -> Metadata.Position.Coords: all fields are now `double` instead of `float`
* in `preset` schema, `icon` was renamed to `iconId` to match proto. We can rename on encoding/decoding conversion functions but it seems kinda cumberstone (i.e. on decoding, this means setting the return object to a `const` and then `delete rest.iconId` so that there's no extra field)

tomasciccola:
Additionally on the top of `test/fixture.js` I cached a bunch of values so we can actually compare docs with some expected result (mostly randomly generated Ids and Dates)

tomasciccola:
Some possible shortcomings that I saw when testing docs with extra - nonvalid- fields:
On every field of type `TagValue` we basically allow ONE field only (which can be a `TagPrimitive` or `TagPrimitive[]`). 
If you add one extra field on encoding, there's no guarantee of which field we're gonna get back.
Is this the desired behavior (allowing only one field)? I honestly can't remember...

gmaclennan:
> On every field of type `TagValue` we basically allow ONE field only (which can be a `TagPrimitive` or `TagPrimitive[]`).
> If you add one extra field on encoding, there's no guarantee of which field we're gonna get back.

This is a bug, it's here: https://github.com/digidem/mapeo-schema/blob/master/src/lib/encode-converstions.ts#L148-L155
The reducer is not adding values to the acculmulator.

gmaclennan:
Fix is:

```js
  return Object.keys(tags).reduce<{ [key: string]: TagValue_1 }>((acc, k) => {
    acc[k] = convertTagValue(tags[k])
    return acc
  }, {})
```

Note you can pass a generic to Array.reduce like I do above, rather than trying to add types to the arguments of the callback.

gmaclennan (approved):
I did some fixing up of the tests and corrected some typos, including acurracy -> accuracy in protobuf.

The commit messages of my changes should explain the changes and why. The main aim of the changes is to be really clear about what we are testing and stripping away extra info and manually written stuff that can be subject to typos.

Can you open a separate issue for the tests that still need to be done? I think we can keep them all in one issue for now.

One change I made: I made all "good docs" be of the correct type, e.g. all fields required by JSON Schema are included. I know our protobuf definitions are more flexible - more fields are optional - but we should test this separately by directly using the protobuf encode methods with minimal fields and then decoding with our `decode()`.

PR #122, opened 2023-08-16
chore: validate schemas to ensure no top-level null

### LINKED ISSUE #104: Prevent schemas from using nullable type for top-level field definitions
We currently do not define any top-level schema fields as nullable because Protobuf doesn't have a proper representation of `null`. However, there's a misalignment between this and how SQLite works, where there's no `undefined` / void column type in SQLite and so optional fields are stored and read as `null`. In Mapeo Core, after reading values from SQLite (via Drizzle), we update fields with `null` values to `undefined` to match the Mapeo Schema definition of an optional field ([link](https://github.com/digidem/mapeo-core-next/pull/146/files#diff-c57734d4459da7c8db888525e982683f635f3641e647f5d59670876d35a26a25R199)).

However, there's currently no protection against us defining a nullable top-level field in the schema, which will cause an issue at the SQLite-to-MapeoCore level because we will replace an _intentional_ `null` value with `undefined` as opposed to leaving it untouched. This does not affect schema fields with a nested JSON-like value definition since we store the JSON as-is in SQLite.

Recommendations:

- add a test to prevent maintainers from defining a top-level nullable field on a schema
- update the schema generation implementation to warn/error if it detects a change that introduces this

gmaclennan:
FIxes #104

### DIFF
--- a/scripts/generate.js
+++ b/scripts/generate.js
@@ -2,13 +2,13 @@
 import fs from 'fs'
 import path from 'path'
 import { mkdirp } from 'mkdirp'
-import { execSync } from 'child_process'
 
 import { parseConfig } from './lib/parse-config.js'
 import { generateProtoTypes } from './lib/generate-proto-types.js'
 import { PROJECT_ROOT } from './lib/utils.js'
 import { generateConfig } from './lib/generate-config.js'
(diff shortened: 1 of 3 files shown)


achou11 (approved):
noice!

Issue #123, opened 2023-08-17
Add tests for encoding/decoding header on mapeo docs


tomasciccola:
We need to add tests for encoding and decoding the header (which now includes only the schemaType as a random generated id and the schema version).
There's the discussion of adding `createdAt` and `createdBy` to the header (see https://github.com/digidem/mapeo-core-next/issues/153), so if this ends up implemented, we should also test for that

gmaclennan:
> There's the discussion of adding `createdAt` and `createdBy` to the header (see [digidem/mapeo-core-next#153](https://github.com/digidem/mapeo-core-next/issues/153)), so if this ends up implemented, we should also test for that

I didn't intend any of the options I described to mean to add to the header. I think these belong in the main block with everything else.

tomasciccola:
oohh, missunderstanding from my part then!

PR #130, opened 2023-08-21
feat: Export decodeBlockPrefix

### LINKED ISSUE #121: Add decode function that only decodes the datatype
Related to https://github.com/digidem/mapeo-core-next/issues/167

For indexing project info, we need to know if the hypercore entry we're indexing is related to project info (we handle project info records differently than other datatypes). Ideally this does not require decoding the whole protobuf value because it's more expensive than necessary. It should be enough to just read the first few bytes to determine the data type and return that in this function.

gmaclennan:
Fixes #121

Lacking tests, tracking in #123. I think ok to merge without tests and add tests later in #123.

### DIFF
--- a/src/decode.ts
+++ b/src/decode.ts
@@ -16,7 +16,7 @@ import {
   convertPreset,
   convertRole,
   convertDevice,
-  convertCoreOwnership
+  convertCoreOwnership,
 } from './lib/decode-conversions.js'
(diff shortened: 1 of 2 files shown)


achou11 (approved):
ah didn't realize this was already implemented - would've done it myself had I known 😄

PR #133, opened 2023-08-22
make project schema properties optional

sethvincent:
This removes `name` and `defaultPresets` from the required array of the project jsonschemas.

There's one type error that I've ts-ignored that I'm not sure how to deal with. See the comment below.

### DIFF
--- a/src/lib/encode-converstions.ts
+++ b/src/lib/encode-converstions.ts
@@ -21,28 +21,29 @@ type ConvertFunction<TSchemaName extends SchemaName> = (
   >
 ) => CurrentProtoTypes[TSchemaName]
 
-/* @ts-ignore TODO: resolve "not assignable to type 'ConvertFunction<"project">" */
 export const convertProject: ConvertFunction<'project'> = (mapeoDoc) => {
   return {
     common: convertCommon(mapeoDoc),
     ...mapeoDoc,
-    defaultPresets: {
-      point: (mapeoDoc.defaultPresets?.point || []).map((p) =>
-        Buffer.from(p, 'hex')
-      ),
-      area: (mapeoDoc.defaultPresets?.area || []).map((a) =>
-        Buffer.from(a, 'hex')
-      ),
-      vertex: (mapeoDoc.defaultPresets?.vertex || []).map((v) =>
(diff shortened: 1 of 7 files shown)


src/lib/encode-converstions.ts
```diff
@@ -21,24 +21,25 @@ type ConvertFunction<TSchemaName extends SchemaName> = (
   >
 ) => CurrentProtoTypes[TSchemaName]
 
+/* @ts-ignore TODO: resolve "not assignable to type 'ConvertFunction<"project">" */
 export const convertProject: ConvertFunction<'project'> = (mapeoDoc) => {
   return {
     common: convertCommon(mapeoDoc),
     ...mapeoDoc,
```
tomasciccola, line 28:
```suggestion
    ...mapeoDoc,
    name: mapeoDoc.name || '',
```
I think that since `name` is now optional, we need to take that into account here

gmaclennan: I changed the protobuf definition to allow for optional values

gmaclennan: I changed the protobuf definition to allow for optional values

src/lib/encode-converstions.ts
```diff
@@ -21,24 +21,25 @@ type ConvertFunction<TSchemaName extends SchemaName> = (
   >
 ) => CurrentProtoTypes[TSchemaName]
 
+/* @ts-ignore TODO: resolve "not assignable to type 'ConvertFunction<"project">" */
```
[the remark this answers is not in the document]
gmaclennan: The types from protobuf are quite confusing. Every field in protobuf is optional, but the decoder defaults to settings undefined fields as an empty string / false / 0, unless fields are marked `optional`. 

Fixed this by defining `name` and `defaultPresets` as optional in the proto file.

gmaclennan (approved):
Made some changes, updated the test, and a fix for the type issue. Looks good to me.

src/lib/encode-converstions.ts
tomasciccola: right! forgot about that!

PR #138, opened 2023-08-24
feat: add optional `deleted` field to common (proto and schema)

### LINKED ISSUE #84: Add `deleted` field to common props
Should be optional. Should only exist if `deleted: true`, otherwise no field should be present. Should be "common" field.

```ts
type Common = {
  // ...
  deleted?: true
}
```

tomasciccola:
this would close #84


tomasciccola:
I'm wondering how to test an invalid value for the `deleted` field. On the JSONSchema level we can force `deleted` to be true but I don't know a way to do it on the protobuf level (maybe with an `option` ??) @gmaclennan

gmaclennan (changes requested):
I suggest we make `deleted` required, it will make all the code easier I think.

gmaclennan (changes requested):
For completeness sake there should be a test for when deleted = `true`

gmaclennan (approved):

PR #139, opened 2023-08-28
feat: header tests

### LINKED ISSUE #123: Add tests for encoding/decoding header on mapeo docs

tomasciccola:
This adds some tests for encoding the headers of a `MapeoDoc`.
Additionally I've added an export for `encodingBlockPrefix` only to be accessible from the unit test.

This should close #123

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -70,7 +70,6 @@ then decoding and comparing the two objects - extra values shouldn't be present`
 })
 
 test(`testing decoding of header that should match the dataTypeId and version`, async (t) => {
-  t.plan(Object.values(dataTypeIds).length * 2)
   for (const [schemaName, dataTypeId] of Object.entries(dataTypeIds)) {
     // TODO: test also schemaVersions greater than the current, for foward compat
     const schemaVersion = currentSchemaVersions[schemaName]
@@ -91,7 +90,6 @@ test(`testing decoding of header that should match the dataTypeId and version`, 
 })
 
 test(`test encoding and decoding of block prefix, ignoring data that comes after`, async (t) => {
-  t.plan(Object.keys(currentSchemaVersions).length * 2)
   for (let [schemaName, schemaVersion] of Object.entries(
     currentSchemaVersions
   )) {
@@ -117,7 +115,6 @@ test(`test encoding and decoding of block prefix, ignoring data that comes after
 })
 
 test(`test encoding of wrongly formatted header`, async (t) => {
-  t.plan(4)
   /** @type { import('../src/types.js').ValidSchemaDef } */
   let schemaDef = {
     schemaName: 'presot',
--- a/test/index.js
+++ b/test/index.js
@@ -15,7 +15,6 @@ import {
   goodDocsCompleted,
   badDocs,
 } from './fixtures/index.js'
-import { throws } from 'node:assert'
 
 test('Bad docs throw when encoding', (t) => {
   for (const { text, doc } of badDocs) {
@@ -156,7 +155,7 @@ test(`test encoding of wrongly formatted header`, async (t) => {
   // hardcoded since there's a slim chance we produce a correct header
   const unknownDataTypeId = Buffer.from('7a79b8b773b2', 'hex')
   unknownDataTypeId.copy(buf)
-  throws(() => {
+  t.throws(() => {
     decodeBlockPrefix(buf)
   })
 })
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -61,6 +68,51 @@ then decoding and comparing the two objects - extra values shouldn't be present`
   }
 })
 
+test(`testing decoding of header that should match the dataTypeId and version`, async (t) => {
+  for (const { doc } of goodDocsCompleted) {
```
gmaclennan, line 72: Maybe use the dataTypeIds map to get the schema names for encoding here, and test a couple few schema versions e.g. 

Note: checking the code, this will actually fail right now, because there is an assert on the schema version. That's a bug - we need this module to be able to parse future schema versions.

```
for (const [schemaName, dataTypeId] of Object.values(dataTypeIds)) {
  for (let schemaVersion = 0; schemaVersion < 128; schemaVersion ++) {
    const buf = encodeBlockPrefix({ schemaName, schemaVersion })
    t.deepEqual(decodeBlockPrefix(buf), { schemaName, schemaVersion })
    t.equal(buf.subarray(0, 6).toString('hex'), dataTypeId)
  }
}
```

Also, testing that the first 6 bytes of the prefix is the data type id, without just re-implementing the compact-encoding code from decodeBlockPrefix.

gmaclennan (changes requested):
I think these tests still need a bit of work.

For encode/decode of block prefixes that work, we are testing most of the functionality in the tests for good docs. What we aren't testing here that would be good to unit test for is different schema versions, e.g. 

```
const schemaDef = { schemaName: 'project', schemaVersion: 5 }
const buf = encodeBlockPrefix(schemaDef)
t.deepEqual(decodeBlockPrefix(buf), schemaDef)
```

Although see my test suggestion in the comment for a more thorough version of the above.

It might also be worth testing (to check it ignores data beyond the block prefix):

```
const schemaDef = { schemaName: 'project', schemaVersion: 5 }
const buf = Buffer.concat(encodeBlockPrefix(schemaDef). randomBytes(50)])
t.deepEqual(decodeBlockPrefix(buf), schemaDef)
```

Finally worth testing we can encode/decode a schema of version `Math.pow(2, 16) -1`

And we should check the length of the encoded block prefix is always 8 in the tests above.


For testing when we expect encode or decode to throw:

1. Calling encodeBlockPrefix with an invalid schemaName
2. Calling decodeBlockPrefix with a Buffer that is less than the prefix length (8 bytes), e.g.

```
const schemaDef = { schemaName: 'project', schemaVersion: 5 }
const buf = encodeBlockPrefix(schemaDef).subarray(0, 7)
throws(() => decodeBlockPrefix(buf))
```

3. Calling decodeBlockPrefix with random data
4. Calling decodeBlockPrefix with an empty buffer
5. Calling decodeBlockPrefix with an unknown data type id e.g.

```
const schemaDef = { schemaName: 'project', schemaVersion: 5 }
const buf = encodeBlockPrefix(schemaDef)
const unknownDataTypeId = randomBytes(6) // Maybe hard-code, or this will fail with a probability of 7/2^6 
unknownDataTypeId.copy(buf)
throws(() => decodeBlockPrefix(buf))
```

That's off the top of my head, any other ideas?

tomasciccola:
> Finally worth testing we can encode/decode a schema of version `Math.pow(2, 16) -1`

whats the reason for this? Is this the biggest `schemaVersion` we should support? 
This would also fail since we're not actually supporting bigger `schemaVersion` than the currently supported, right??

gmaclennan:
> whats the reason for this? Is this the biggest `schemaVersion` we should support?
> This would also fail since we're not actually supporting bigger `schemaVersion` than the currently supported, right??

Because what we think we have done is use 2 bytes (16 bits) to store the schemaVersion - we had that conversation a while ago. We could have chosen 1 byte (schema version up to 256) but we chose 2 bytes to be safe - probably overkill now I think about it! Anyway, given that is what we think we are doing, I thought it's worth testing that is actually the case. This will fail until we fix the bug that does not allow a schemaVersion bigger than currently supported.

gmaclennan (approved):
looks good, good work.

PR #140, opened 2023-08-31
add hash in attachments to observation schema

sethvincent:
A change to the observation schema to support digidem/mapeo-core-next#220.


achou11 (approved):
not sure i should have the final say on this but changes lgtm

gmaclennan (changes requested):
Can we do the same with this as we do with other fields that are buffers, and store in the protobuf as bytes, and stringify in encode.

gmaclennan (approved):

PR #143, opened 2023-09-12
chore: fix misspelling in filename

achou11:
just bothersome enough when doing file searches 😅


gmaclennan (approved):
yeah been meaning to fix this.

PR #147, opened 2023-09-26
chore!: rename Attachment.driveId to Attachment.driveDiscoveryId

achou11:
Towards https://github.com/digidem/mapeo-core-next/issues/272

My understanding from the issue is that we want to rename this, but not entirely sure 😄


gmaclennan (approved):
Yes was going to say name it driveDiscoverId not discoveryDriveId but I see the former is what you use in the code 👍

tomasciccola (approved):

PR #156, opened 2023-10-09
fix: iconVersionId wrong encode decode

tomasciccola:
on `src/lib/{encode,decode}-conversions.js` when converting the `iconVersionId` for `IconVariant` we where encoding//decoding the string as 'hex'. 
Since the strings has a slash ('/'), we where losing it when encoding the document. 
Tests weren't failing since we weren't generating a valid `versionId` (without the `/${index}`)

- [x] Check if doing `Buffer.from(string, 'utf-8')` and `buf.toString('utf-8')` is safe

### DIFF
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -208,7 +208,7 @@ function convertIconVariant(variant: Icon_1_IconVariant) {
   const { blobVersionId, mimeType, size, pixelDensity } = variant
   return {
     blobVersionId: blobVersionId.toString('hex'),
-    mimeType: convertIconMimeType(mimeType) as ValidMimeTypes,
+    mimeType: convertIconMimeType(mimeType),
     size: size === 'UNRECOGNIZED' ? 'medium' : size,
     pixelDensity: convertIconPixelDensity(pixelDensity),
   }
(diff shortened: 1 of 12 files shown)


achou11:
Not super sure about the changes here. Makes me wonder if the `blobVersionId` field should be derived similarly to how we derive `versionId` in the Common schema.

That has a `links` field, where a `Link` is `{ coreDiscoveryKey: string, index: number }`

tomasciccola:
> Makes me wonder if the `blobVersionId` field should be derived similarly to how we derive `versionId` in the Common schema.

Yeah, I think that makes a lot of sense. I made it those changes on the last commit but now I have a type error because `Icon.blobVersionId` is typed as optional on the protobuf typing. 
I thought putting `[(required) = true]` makes the field required (in some sense...). Is there smth that I'm missing? 'cause giving that field a default value doesn't make much sense... @gmaclennan any pointers?

src/lib/decode-conversions.ts
```diff
@@ -207,7 +207,7 @@ export const convertIcon: ConvertFunction<'icon'> = (message, versionObj) => {
 function convertIconVariant(variant: Icon_1_IconVariant) {
   const { blobVersionId, mimeType, size, pixelDensity } = variant
```
gmaclennan, line 208:
```suggestion
  const { blobVersionId, mimeType, size, pixelDensity } = variant
    if (!blobVersionId) {
    throw new Error('Missing required property `blobVersionId`')
  }
```

gmaclennan (approved):

PR #164, opened 2024-01-31
feat: add optional deviceType to deviceInfo

gmaclennan:
Towards https://github.com/digidem/mapeo-core-next/issues/455

### DIFF
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -13,5 +13,12 @@ message DeviceInfo_1 {
 
   Common_1 common = 1;
 
+  enum DeviceType {
+    mobile = 0;
+    tablet = 1;
+    desktop = 2;
+  }
+
   string name = 5;
+  optional DeviceType deviceType = 6;
 }
--- a/test/fixtures/good-docs-completed.js
+++ b/test/fixtures/good-docs-completed.js
@@ -170,9 +170,28 @@ export const goodDocsCompleted = [
       createdBy: cachedValues.createdBy,
       links: [],
       name: 'my device name',
+      deviceType: 'desktop',
       deleted: true,
     },
     expected: {},
+  },
+  {
+    doc: {
+      docId: cachedValues.docId,
+      versionId: cachedValues.versionId,
+      schemaName: 'deviceInfo',
+      createdAt: cachedValues.createdAt,
+      updatedAt: cachedValues.updatedAt,
+      createdBy: cachedValues.createdBy,
+      links: [],
+      name: 'my device name',
+      // @ts-expect-error
+      deviceType: 'motorbike',
+      deleted: true,
+    },
+    expected: {
+      deviceType: 'UNRECOGNIZED',
+    },
   },
   {
     doc: {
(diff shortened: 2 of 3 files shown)


test/fixtures/good-docs-completed.js
```diff
@@ -170,10 +170,29 @@ export const goodDocsCompleted = [
...
+      docId: cachedValues.docId,
+      versionId: cachedValues.versionId,
+      schemaName: 'deviceInfo',
+      createdAt: cachedValues.createdAt,
+      updatedAt: cachedValues.updatedAt,
+      createdBy: cachedValues.createdBy,
+      links: [],
+      name: 'my device name',
+      // @ts-expect-error
+      deviceType: 'motorbike',
```
EvanHahn, line 189: Could we test what happens if `deviceType` is missing?

proto/deviceInfo/v1.proto
```diff
@@ -13,5 +13,12 @@ message DeviceInfo_1 {
 
   Common_1 common = 1;
 
+  enum DeviceType {
+    mobile = 0;
```
EvanHahn, line 17: [In the past][0], I've seen `0` reserved for the unknown case. If the field is completely missing, that enum case gets used. Is that a problem for us here?

(This is at the edge of my protobuf knowledge so I might be totally wrong.)

[0]: https://github.com/signalapp/Signal-Desktop/blob/78fbd8c8ceabbda6b15f77084e966d84dc96ce8d/protos/SignalService.proto#L12

EvanHahn (approved):
LGTM assuming my enum question isn't a problem.

proto/deviceInfo/v1.proto
gmaclennan: As far as I understand it, at least with the protobuf parser that we use, if it encounters an enum that is not defined (eg in this case not 0,1,2) then it parses as "UNRECOGNISED". So I think protobuf and/or the parser already do what you are looking for. I think you are suggesting an addition enum definition `unknown = 0` but that would result in two different values to deal with - "unknown" and "UNRECOGNISED". We could just not define the `0` field and start with `1` - I don't see a problem with that, it maybe avoids mis-parsing empty data (`0`) as meaning something.

gmaclennan: Sorry just checked you link, and I think this might be more for needing an enum for "the thing this enum represents is unknown" rather than "the enum being parsed is unrecognized". I decided to make this an explicit optional field so that it being undefined holds the meaning of "we don't know the device type, or the device chose not to share it". Then for future compatibility, a new device type eg `4 = VR_headset` would show as `UNRECOGNISED ` to older clients

EvanHahn: That makes sense. As long as an unknown value (eg `4`) and a missing value are both handled as expected, I'm happy.

(I'm happy either way. My happiness in life does not depend on this.)

test/fixtures/good-docs-completed.js
gmaclennan: This is already tested here: https://github.com/digidem/mapeo-schema/blob/ad39db8ab5dc5ca4c3211bb212157c0f8f151b85/test/fixtures/good-docs-minimal.js#L97-L107

PR #166, opened 2024-02-06
feat: translation

### LINKED ISSUE #165: Translation record type
Rough schema:

```protobuf
message Translation {
  recordType: string // lets say, 'field' or 'role'
  recordId: byte[] // the id that points to the record we want to get a translation for
  fieldRef: string[] // the field from that record that we want to translate the string for
  languageCode: string // BCP-47 or ISO 638-3
  message: string // the actual translated string
 
}
```

tomasciccola:
should close #165


tomasciccola:
wondering how to validate `languageCode`. I know js has `Intl.Locale` but it seems you can pass some arbitrary stuff and it won't throw. I found [this](https://www.npmjs.com/package/bcp47) this library last updated 9 years ago that doesn't seem to do what it claims (return null on a valid bcp47 string)

tomasciccola:
another nice to have would be to validate `recordType` (check if the schema exists) and `fieldRef`,  traversing the array and checking if the provided field exists for that recordType.
validating `recordId` I don't think its easy since it may point to a record we don't know it exists yet?

tomasciccola:
Another thing to address is how to solve a field ref for a field that is in an Array. i.e. 
The `options` field in the `Field` record type is an array of `{label:string, value: tagValue}`
A `fieldRef` pointing to that would be smth like `['options',0, 'label']` ? meaning, 'this points to the 'label' field in the first option of a field??

tomasciccola:
> wondering how to validate `languageCode`. I know js has `Intl.Locale` but it seems you can pass some arbitrary stuff and it won't throw. I found [this](https://www.npmjs.com/package/bcp47) this library last updated 9 years ago that doesn't seem to do what it claims (return null on a valid bcp47 string)

I went ahead and installed [is-language-code](https://www.npmjs.com/package/is-language-code) which seems supported and up to date, but it uses smth called [iana](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry) for language codes, which I don't know if we want to...

gmaclennan:
I don't think we should validate when encoding / decoding. I think we should do that when writing. Invalid data will just be ignored, because lookups will just not work. If possible, we can add a type to `recordType` (`Exclude<MapeoDoc['schemaName'], 'translation'>`).

I wrote up thoughts on language codes [in the issue](https://github.com/digidem/mapeo-schema/issues/165#issuecomment-1933982898).

For the fieldRef, I think we should just use a string, using [dot-prop notation](https://github.com/sindresorhus/dot-prop). This solves the question about field options (the example would be `options[0].label`). I don't think the encoding or decoding here should do anything with that prop though.

package.json
```diff
@@ -55,6 +55,7 @@
     "ajv-formats": "^2.1.1",
     "compact-encoding": "^2.12.0",
     "glob": "^10.3.3",
+    "is-language-code": "^5.0.12",
```
gmaclennan, line 58: remove, I think.

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
...
+      "type": "string",
+      "description": "type of record related to the field being translated"
+    },
+    "recordId": {
+      "type": "string",
+      "description": "the id that points to the record we want to get a translation for"
+    },
+    "fieldRef": {
+      "type": "string",
+      "description": "the field from that record that we want to translate"
```
gmaclennan, line 22:
```suggestion
      "description": "identifier for translated field/property in dot-prop notation"
```

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
...
+  "title": "translation",
+  "type": "object",
+  "properties":{
+    "schemaName":{
+      "type":"string",
+      "const": "translation"
+    },
+    "recordType": {
+      "type": "string",
+      "description": "type of record related to the field being translated"
```
gmaclennan, line 14:
```suggestion
      "description": "schema name of record related to the field being translated"
```

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "http://mapeo.world/schemas/translation/v1.json",
+  "description": "A translation is a value for a string - field, preset, tag - that is translated at runtime",
```
gmaclennan, line 4:
```suggestion
  "description": "A translation is a translated message in a single language for a string property/field of any Mapeo record type",
```

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
...
+      "type":"string",
+      "const": "translation"
+    },
+    "recordType": {
+      "type": "string",
+      "description": "type of record related to the field being translated"
+    },
+    "recordId": {
+      "type": "string",
+      "description": "the id that points to the record we want to get a translation for"
```
gmaclennan, line 18:
```suggestion
      "description": "the docId of the record that this field is translated for"
```

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
...
+    },
+    "fieldRef": {
+      "type": "string",
+      "description": "the field from that record that we want to translate"
+    },
+    "languageCode": {
+      "type": "string",
+      "description": "code that identifies the language we are translated for (for now BCP-47)"
```
gmaclennan, line 26:
```suggestion
      "description": "three-letter ISO 169-3 language code"
```

schema/translation/v1.json
```diff
@@ -0,0 +1,35 @@
...
+    },
```
gmaclennan, line 27:
```suggestion
    },
    "regionCode": {
        "type": "string",
        "description": "two-letter country code from ISO 3166-1 alpha-2 or a three-digit code from UN M.49 for geographical regions"
},
```

proto/translation/v1.proto
```diff
@@ -0,0 +1,20 @@
...
+  option (dataTypeId) = "cb2327c9a27f";
+  option (schemaName) = "translation";
+
+  Common_1 common = 1;
+
+  string recordType = 2;
+  bytes recordId = 3;
+  string fieldRef = 4;
+  string languageCode = 5;
+  string message = 6;
```
gmaclennan, line 18:
```suggestion
  string languageCode = 5;
  string regionCode = 6;
  string message = 7;
```

gmaclennan (approved):
Just missing the region code, plus some suggested copy edits. Otherwise good to merge I think.

proto/translation/v1.proto
```diff
@@ -0,0 +1,20 @@
...
+
+message Translation_1 {
+  // **DO NOT CHANGE dataTypeId** generated with `openssl rand -hex 6`
+  option (dataTypeId) = "cb2327c9a27f";
+  option (schemaName) = "translation";
+
+  Common_1 common = 1;
+
+  string recordType = 2;
+  bytes recordId = 3;
```
gmaclennan, line 15:
```suggestion
  string schemaNameRef = 2;
  bytes docIdRef = 3;
```

Thinking maybe rename these for more consistency?

PR #172, opened 2024-02-19
chore!: rename role -> membership

gmaclennan:
BREAKING: Fixes Suggestion: rename `role` to `membership` #163

### DIFF
--- a/proto/membership/v1.proto
+++ b/proto/membership/v1.proto
@@ -6,10 +6,10 @@ import "google/protobuf/timestamp.proto";
 import "common/v1.proto";
 import "options.proto";
 
-message Role_1 {
+message Membership_1 {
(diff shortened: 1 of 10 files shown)


EvanHahn (approved):

EvanHahn:
Let me know if you want me to take another look at this.

gmaclennan:
@EvanHahn Do we want to maybe do this before v1 (e.g. before MVP), simply to avoid a breaking change post-mvp? I realize it's confusing - this doesn't change the encoded proto format, but it is a breaking change for the decoded data because of the rename.

EvanHahn:
This is a good change but I don't think we should delay the MVP, even though it's probably a small amount of work to integrate. Personally, I would punt this to post-MVP.

gmaclennan:
Agreed.

PR #173, opened 2024-02-22
feat: two icon types

### LINKED ISSUE #160: [icon] should a variant's pixelDensity only be defined for applicable mime types?
Surfaced during https://github.com/digidem/mapeo-core-next/pull/335, but `pixelDensity` doesn't really have any impact on svg assets and only applies for bitmap formats such as jpeg and png. as [currently defined](https://github.com/digidem/mapeo-schema/blob/afa83dff7266f5d61d2a89193bb9d8de5926447f/schema/icon/v1.json), that field is non-optional and expects a valid enum value. ideally that field only exists for the applicable mimetype via a discriminate union, but doesn't seem like json schema truly supports that. looks like you'd have to fully write out the different variants and use `anyOf`, which may be okay but it's definitely verbose.

maybe this is a non-problem in general and applications just have to understand that the pixel density is ideally inconsequential for svg variants.

tomasciccola:
this should fix #160

### DIFF
--- a/proto/icon/v1.proto
+++ b/proto/icon/v1.proto
@@ -1,0 +1,64 @@
+syntax = "proto3";
+package mapeo;
+
+import "common/v1.proto";
+import "options.proto";
+
+message Icon_1 {
+  // **DO NOT CHANGE dataTypeId** generated with `openssl rand -hex 6`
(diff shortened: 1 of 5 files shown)


tomasciccola:
closed because git screw up :'|

PR #178, opened 2024-04-08
feat: Expose validation errors

gmaclennan:
Exposes validation errors as an array on the `validate()` function, like Ajv does.

Also adds an export of the `valueOf` utility, which can replace the version used in mapeo-core-next.


tomasciccola (approved):
LGTM! thanks! :100:

PR #179, opened 2024-04-16
feat: Add track schema type

gmaclennan:
Add a new `track` schema type.

This was more work than I would have hoped because I wanted to keep timestamps as integers for track locations (as opposed to converting them to ISO date strings), which required lots of backwards compat work. This was maybe a bad idea. But maybe we should switch to timestamps instead of date strings throughout, or switch to `Date` (which would require custom revivers for serializing as JSON over IPC).

Currently this is storing all data captured during track recording, apart from `altitude accuracy`, which seems unnecssary.

To display a track we only really need the latitude and longitude of each position along a track, but at the very least the timestamp is useful too, and the speed, altitude, accuracy and heading could be useful in the future. Each value adds approx 4 bytes to the stored data (64 bits). According to the SWM team tracks capture about 333 points per 1,000 meters, which, if users are going to walk an average of 10km on a track, means an overhead of 333 * 10 * 4 = ~13KB storage cost for each track per value stored.

Note: at some time soon (before adding another schema type?) it would be good to write more generic encode and decode functions, based on annotations of the protobufs maybe.

TODO:

- [x] Add tests for track encoding and decoding.

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -9,7 +9,7 @@ plugins:
       - importSuffix=.js
       - env=node
       - snakeToCamel=false
-      - useDate=string
+      - useDate=false
       # - useOptionals=none
       - stringEnums=true
       - enumsAsLiterals=true
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -9,7 +9,7 @@ plugins:
       - importSuffix=.js
       - env=node
       - snakeToCamel=false
-      - useDate=false
+      - useDate=string
       # - useOptionals=none
       - stringEnums=true
       - enumsAsLiterals=true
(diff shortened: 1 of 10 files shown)


EvanHahn:
Made a few small changes:

- Removed a redundant check: d2912f5b642fbc33f71f00b43d069ee0b23ab5ff
- Tweaked documentation: 015fcc5a088a793e3cb9b2fe42052cd879348946
- Added test fixtures: 40270f3186e4181a05844f068e17d42a329c7fe5

I think this is now ready to merge.

EvanHahn (approved):

EvanHahn:
I'll release this shortly.

PR #183, opened 2024-05-14
feat: add color to preset

### LINKED ISSUE #182: Add "color" field to Presets
On configs presets have a color field that is used for icon borders and observation point colors

tomasciccola:
should fix #182 

Validation is done at the JSONSchema level by using a regex pattern in the schema itself (that only allows colors in format #rrggbb) and before encoding by using the [validate-color](https://www.npmjs.com/package/validate-color) lib. 

Looking at the default config, I see there are some presets that have no color. How should we handle that? A possibility is to add a default color on the config-builder (I don't think that should live here but maybe it should??)


schema/preset/v2.json
```diff
@@ -93,17 +93,23 @@
       "items": {
         "type": "string"
       }
+    },
+    "color": {
+      "description": "string representation of a color in 24 bit (#rrggbb)",
+      "type": "string",
+      "pattern": "^#([a-fA-F0-9]{6})$"
```
EvanHahn, line 100: Nitpick: I don't think we need this [capturing group](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Capturing_group).

```suggestion
      "pattern": "^#[a-fA-F0-9]{6}$"
```

src/lib/encode-conversions.ts
```diff
@@ -66,6 +67,9 @@ export const convertField: ConvertFunction<'field'> = (mapeoDoc) => {
 }
 
 export const convertPreset: ConvertFunction<'preset'> = (mapeoDoc) => {
+  if (!validateColor.validateHTMLColorHex(mapeoDoc.color)) {
```
EvanHahn, line 70: I worry that this will get out of sync with our regex. Could we construct the regex from what we have in `schema/preset/v2.json`?

EvanHahn (approved):
> wondering: should we validate when encoding that a color value is a 24-bit color string? for now, validation is done on JSONSchema level by adding a `pattern` with a regex in the schema itself. This means that passing an invalid string would throw. I wonder how we should handle that? 
> just let it throw?
> Should we have a default color as a fallback? Should this default color be handled here? It could also live on the settings-builder (basically if no color or invalid color lives on a preset, then add the default one...)

IMO, we should let it throw here because the color is invalid. The settings builder seems like a great place to use a fallback.

src/lib/encode-conversions.ts
tomasciccola: done on the last commit!

schema/preset/v2.json
tomasciccola: Yeah, it was left from when it also mached against 12-bit color (so, like #rgb)

PR #190, opened 2024-07-02
feat: Remove old schemas and re-version all schemas and protobufs to v1

### LINKED ISSUE #189: Move old schemas to new repo, re-version all schemas to v1
we should create a new repo ([i.e. @mapeo/schema-legacy](https://github.com/digidem/schema-legacy)) and move all the schemas on `old/` to that repo.
additionally we should change all version (through file renaming) to v1 before MPV. This versioning change wouldn't be a breaking change since we always use the last version for validation and encoding

tomasciccola:
Should close #189


proto/projectSettings/v1.proto
```diff
@@ -5,6 +5,8 @@ import "google/protobuf/timestamp.proto";
 import "common/v1.proto";
 import "options.proto";
 
+// Same as v1, created to test forward compatibility
+
```
EvanHahn, line 9: I don't understand this comment. Should we remove it?

```suggestion
```

EvanHahn (approved):

proto/projectSettings/v1.proto
tomasciccola: projectSettings was copied (with the same content) to test forward compatibility. It doesn't make sense to keep it now...

tomasciccola:
I'm wondering what work is needed left for bumping this to a major version (v1). Looking at the release.yml I see that `prerelease` is publish under `next` dist-tag, but this release would be a major version so I'm wondering if the release flow would take care of this...

tomasciccola:
I'm going to go ahead and merge it. We can address my comment later

EvanHahn:
Yeah, #371 is a breaking change that would require a major version. I think we should wait until that's done.

PR #191, opened 2024-07-10
feat!: replace `createdBy` with `originalVersionId`

EvanHahn:
See <https://github.com/digidem/mapeo-core-next/issues/371> for details.


proto/common/v1.proto
```diff
@@ -14,8 +14,7 @@ message Common_1 {
   repeated Link links = 2;
   google.protobuf.Timestamp createdAt = 3 [(required) = true];
   google.protobuf.Timestamp updatedAt = 4 [(required) = true];
-  // 32-byte hash of the discovery key of a core
-  bytes createdBy = 5 [(required) = true];
+  optional string originalVersionId = 5;
```
gmaclennan, line 17: This should be the same type as `Link`:

```
message Link {
    bytes coreDiscoveryKey = 1;
    int32 index = 2;
  }
```

Maybe rename `Link -> VersionId`

schema/common/v1.json
```diff
@@ -51,7 +51,7 @@
     "updatedAt",
     "links",
     "versionId",
-    "createdBy",
+    "originalVersionId",
```
gmaclennan, line 54: This does get confusing, because is this required or not? It seems from https://github.com/digidem/mapeo-core-next/pull/715 that we are not storing this on creation, so reads will not include this/

Trying to fully represent the type in JSONSchema might be difficult, it might be easier to do in the TypeScript defs in this repo, maybe something like:

```ts
type CommonNew = (Omit<Common, 'links' | 'originalVersionId'> & { links: [] }) | SetRequired<Common, 'originalVersionId'>
```

e.g. you originalVersionId can be missing if link is an empty array, but otherwise it is required.

src/encode.ts
```diff
@@ -29,7 +29,7 @@ import { ExhaustivenessError } from './lib/utils.js'
  * with the encoded data type ID and schema version, to send to an hypercore.
  */
 export function encode(
-  mapeoDoc: OmitUnion<MapeoDocInternal, 'versionId'>
+  mapeoDoc: OmitUnion<MapeoDocInternal, 'versionId' | 'originalVersionId'>
```
gmaclennan, line 32: I think it's optional rather than omitted?

src/lib/decode-conversions.ts
```diff
@@ -346,13 +346,24 @@ function convertCommon(
...
+
+  /** @type {string} */ let originalVersionId
+  if (common.originalVersionId) {
+    originalVersionId = common.originalVersionId
+  } else if (common.links.length === 0) {
+    originalVersionId = versionId
+  } else {
+    throw new Error('Cannot determine original version ID; data is malformed')
+  }
+
```
gmaclennan, line 359: Ah, I see re. my comment above about it being required in JSON Schema is maybe ok given you are always appending it here.

gmaclennan: But maybe the encode function could benefit from the stricter typing and this same check when writing.

src/lib/encode-conversions.ts
```diff
@@ -28,7 +29,7 @@ import {
  * schema name, and returning the most recent JSONSchema type */
 type ConvertFunction<TSchemaName extends SchemaName> = (
   mapeoDoc: Extract<
-    OmitUnion<MapeoDocInternal, 'versionId'>,
+    OmitUnion<MapeoDocInternal, 'versionId' | 'originalVersionId'>,
```
gmaclennan, line 32: I think we're loosing some type safety where you are doing this, but it's late so maybe I'm missing something?

gmaclennan (changes requested):
Thanks for this. I think we should encode this like we do the versionIds for Links. Also some questions about the typing of docs with originalVersionId and making sure we're not loosing type safety, particularly on writes.

proto/common/v1.proto
EvanHahn: Done in e3024bea0987037111ec38bf2d1787d3a8972775.

src/encode.ts
EvanHahn: Done in d1b0975b5e2cf8f2beee6a1a2ad5fcc6fd251a92.

src/lib/decode-conversions.ts
EvanHahn: I _think_ I addressed this in d1b0975b5e2cf8f2beee6a1a2ad5fcc6fd251a92.

src/lib/encode-conversions.ts
EvanHahn: I think I did this in d1b0975b5e2cf8f2beee6a1a2ad5fcc6fd251a92, but maybe you were imagining something else.

gmaclennan (approved):
This looks good to merge.

PR #196, opened 2024-07-22
chore!: remove `refs` property from schemas

### LINKED ISSUE #193: Remove `refs` property from observation proto and JSONSchema
This property is not currently used, and when we do introduce it, we will likely want to also include a `type` property, see #192

EvanHahn:
This isn't necessary for now, so we can remove it.

`refs` is still present for `Track`s.

Tested this with automated tests and by integrating it into `@mapeo/core` with a temporary build of this module.

Closes #193.

### DIFF
--- a/proto/observation/v1.proto
+++ b/proto/observation/v1.proto
@@ -17,11 +17,6 @@ message Observation_1 {
   optional double lat = 5;
   optional double lon = 6;
 
-  message Ref {
-    bytes id = 1;
-  }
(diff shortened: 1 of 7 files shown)


tomasciccola:
Aproved, but wondering, why are we keeping refs on tracks?

tomasciccola (approved):

EvanHahn:
I don't know much about tracks so I don't know if they're needed. I'll ask.

PR #197, opened 2024-07-22
feat: add `configMetadata` to `projectSettings`

tomasciccola:
This should move forward exposing config metadata to the frontend. 
Part of https://github.com/digidem/comapeo-mobile/issues/527

### DIFF
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -20,9 +20,13 @@ message ProjectSettings_1 {
     repeated bytes relation = 5;
   };
 
-  optional DefaultPresets defaultPresets = 2;
+  message ConfigMetadata {
+    optional string name = 1;
+    optional google.protobuf.Timestamp buildDate = 2;
+    optional int32 fileVersion = 3;
+  }
 
-  optional string name = 5;
-  string configName = 6;
-  string configVersion = 7;
+  optional DefaultPresets defaultPresets = 2;
+  optional ConfigMetadata configMetadata = 3;
+  optional string name = 4;
 }
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -24,7 +24,7 @@ message ProjectSettings_1 {
     string name = 1;
     google.protobuf.Timestamp buildDate = 2;
     google.protobuf.Timestamp importDate = 3;
-    uint32 fileVersion = 4;
+    string fileVersion = 4;
   }
 
   optional DefaultPresets defaultPresets = 2;
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -23,7 +23,8 @@ message ProjectSettings_1 {
   message ConfigMetadata {
     optional string name = 1;
     optional google.protobuf.Timestamp buildDate = 2;
-    optional int32 fileVersion = 3;
+    optional google.protobuf.Timestamp importDate = 3;
+    optional int32 fileVersion = 4;
   }
 
   optional DefaultPresets defaultPresets = 2;
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -21,10 +21,10 @@ message ProjectSettings_1 {
   };
 
   message ConfigMetadata {
-    optional string name = 1;
-    optional google.protobuf.Timestamp buildDate = 2;
-    optional google.protobuf.Timestamp importDate = 3;
-    optional int32 fileVersion = 4;
+    string name = 1;
+    google.protobuf.Timestamp buildDate = 2;
+    google.protobuf.Timestamp importDate = 3;
+    uint32 fileVersion = 4;
   }
 
   optional DefaultPresets defaultPresets = 2;
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -23,4 +23,6 @@ message ProjectSettings_1 {
   optional DefaultPresets defaultPresets = 2;
 
   optional string name = 5;
+  string configName = 6;
+  string configVersion = 7;
 }
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -24,6 +24,8 @@ import { ExhaustivenessError, VersionIdObject, getVersionId } from './utils.js'
 import type { Observation, Track } from '../index.js'
 import type { Observation_1_Attachment } from '../proto/observation/v1.js'
 import type { Track_1_Position } from '../proto/track/v1.js'
+import { ProjectSettings_1_ConfigMetadata } from '../proto/projectSettings/v1.js'
+import { ProjectSettings } from '../schema/projectSettings.js'
 
 /** Function type for converting a protobuf type of any version for a particular
  * schema name, and returning the most recent JSONSchema type */
@@ -38,6 +40,7 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
 ) => {
   const { common, schemaVersion, defaultPresets, ...rest } = message
   const jsonSchemaCommon = convertCommon(common, versionObj)
+  const configMetadata = convertConfigMetadata(message.configMetadata)
   return {
     ...jsonSchemaCommon,
     ...rest,
@@ -50,8 +53,23 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
           relation: defaultPresets.relation.map((r) => r.toString('hex')),
         }
       : undefined,
-    configMetadata: message.configMetadata || {},
+    configMetadata,
   }
+}
+
+function convertConfigMetadata(
+  configMetadata: ProjectSettings_1_ConfigMetadata | undefined
+): ProjectSettings['configMetadata'] {
+  if (!configMetadata) {
+    throw new Error('Missing required property configMetadata')
+  }
+  if (!configMetadata.importDate) {
+    throw new Error('Missing required property configMetadata.importDate')
+  }
+  if (!configMetadata.buildDate) {
+    throw new Error('Missing required property configMetadata.buildDate')
+  }
+  return configMetadata as ProjectSettings['configMetadata']
 }
 
 export const convertObservation: ConvertFunction<'observation'> = (
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -50,6 +50,7 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
           relation: defaultPresets.relation.map((r) => r.toString('hex')),
         }
       : undefined,
+    configMetadata: message.configMetadata || {},
   }
 }
 
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -40,7 +40,10 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
 ) => {
   const { common, schemaVersion, defaultPresets, ...rest } = message
   const jsonSchemaCommon = convertCommon(common, versionObj)
-  const configMetadata = convertConfigMetadata(message.configMetadata)
+  let configMetadata
+  if (rest.configMetadata) {
+    configMetadata = convertConfigMetadata(rest.configMetadata)
+  }
   return {
     ...jsonSchemaCommon,
     ...rest,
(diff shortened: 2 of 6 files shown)


proto/projectSettings/v1.proto
```diff
@@ -23,4 +23,6 @@ message ProjectSettings_1 {
   optional DefaultPresets defaultPresets = 2;
 
   optional string name = 5;
+  string configName = 6;
+  string configVersion = 7;
```
EvanHahn, line 27: I think this is fine but I wonder if a simple integer would be simpler for users? Two reasons:

1. Semver is probably a little harder for users to understand
2. A `string` could contain garbage data

```suggestion
  uint32 configVersion = 7;
```

EvanHahn (changes requested):
One question about whether `configVersion` should be an integer instead...not sure.

proto/projectSettings/v1.proto
tomasciccola: I agree, but maybe should loop in @gmaclennan  since he's thinking about semversioning configs (see https://github.com/digidem/comapeo-mobile/issues/527)

tomasciccola: An issue that comes with this is: 
`configVersion` comes from the `version` field on the `package.json` which is usually a string; we could:
* find another way of denoting version of a config
* having some mechanism to enforce that the 'version' field is a valid int (maybe on build time of a config)
* keeping it as a string and be prepared to catch possible garbage
* some other that I haven't thought of

I don't favor anyone (maybe more inclined to second option?).

tomasciccola: Also wondering about version in the context of a non-technical person. Does an arbitrary number convey that meaning? Wouldn't a date be a more universal way of expressing versioning (so basically, expressing when this config was built?)

gmaclennan:
Following up here from our call about this today. Proposal is to have these fields:

- `configName: string`: human readable name for the config, defined by the person who creates the config.
- `configBuildDate: string`: ISO date of when the config was built. Should encode to proto as a timestamp, like we do for other ISO date fields.
- `configVersion: string`: We discussed the merits of having a semver version name, but decided to not use it for now due to the challenges for the person creating the config to know how to increment the version. We discussed keeping it for future use, but on reflection I suggest we drop it from the schema for now, and add it if needed later. I've been burned by adding things to the schema that we don't actually use in the past.

Separately, I think it's worth capturing some additional metadata when a config is imported:

- `configImportDate` - the date the config was actually imported into the project, different to the build date.
- `configFileName` - the name of the file that the config was imported from.
- `configFileVersion` - the config file format version

I think this data could be useful for debugging and we could expose it to the user in the future, and at least include it in debug info.

With all this data I think maybe put it all under a property, e.g. `configMetadata`? I realize that's jjust me being OCD about organization...

tomasciccola:
> Following up here from our call about this today. Proposal is to have these fields:

>     * `configVersion: string`: We discussed the merits of having a semver version name, but decided to not use it for now due to the challenges for the person creating the config to know how to increment the version. We discussed keeping it for future use, but on reflection I suggest we drop it from the schema for now, and add it if needed later. I've been burned by adding things to the schema that we don't actually use in the past.

I agree about this and was in fact not adding `configVersion` on the schema. `metadata.json` on config files will still save the `configVersion`, we can just ignore it for now
> 
> 
> Separately, I think it's worth capturing some additional metadata when a config is imported:


>     * `configImportDate` - the date the config was actually imported into the project, different to the build date.
Ok, I'll add that field

>     * `configFileName` - the name of the file that the config was imported from.

Not that sure about the utility of this since we are already adding name, but don't have anny issues with adding it
> 
>     * `configFileVersion` - the config file format version

I'm adding this as an int32. And then in `mapeo-core-next` we could have a `SUPPORTED_CONFIG_VERSIONS = [1,2..]` so that we can check if we're importing a config format we support. Do you think this is enough, or should we set it as a string and use semver? I think the complexities of parsing a semver string may not be worth it, but we can discuss it further...


> With all this data I think maybe put it all under a property, e.g. `configMetadata`? I realize that's jjust me being OCD about organization...

heh, I'm actually doing this (nesting everything under `configMetadata`.

On the other hand, I'm making `configMetadata` and its fields optional since a user may create a project (which means calling `setProjectSettings`) before importing a config.

proto/projectSettings/v1.proto
```diff
@@ -20,7 +20,14 @@ message ProjectSettings_1 {
     repeated bytes relation = 5;
   };
 
-  optional DefaultPresets defaultPresets = 2;
+  message ConfigMetadata {
+    optional string name = 1;
+    optional google.protobuf.Timestamp buildDate = 2;
+    optional google.protobuf.Timestamp importDate = 3;
+    optional int32 fileVersion = 4;
```
EvanHahn, line 27: Nit: if the minimum value is `0`, this should be an `uint32`.

```suggestion
    optional uint32 fileVersion = 4;
```

EvanHahn (approved):

gmaclennan:
> I'm adding this as an int32. And then in `mapeo-core-next` we could have a `SUPPORTED_CONFIG_VERSIONS = [1,2..]` so that we can check if we're importing a config format we support. Do you think this is enough, or should we set it as a string and use semver? I think the complexities of parsing a semver string may not be worth it, but we can discuss it further...

This doesn't support the case of forwards-compatibility. E.g. the version of comapeo we release today needs to know which future versions of the file format it is able to parse. I suggest we use [comver](https://gitlab.com/staltz/comver) for the file format (`MAJOR.MINOR`). The value stored in this field will not be parsed (it's just there to support debugging in the future), but for the file reader we should attempt to parse this and throw if it can't be parsed e.g.

```js
function parseVersion(version) {
    const regex = /^(\d+)\.(\d+)$/;
    const match = version.match(regex);

    if (!match) {
        throw new Error('Invalid file version: ' + version);
    }

    const major = parseInt(match[1], 10);
    const minor = parseInt(match[2], 10);

    if (major > SUPPORTED_VERSION) {
        throw new Error('Invalid version: MAJOR cannot be greater than ' + SUPPORTED_VERSION);
    }

    return {
        major: major,
        minor: minor
    };
}
```

proto/projectSettings/v1.proto
gmaclennan, line 27: I think we should make all of these field required. It will make front-end code easier. The file format can ensure they are required, so we know they will always exist if the configMetadata property exists.

gmaclennan (changes requested):

proto/projectSettings/v1.proto
tomasciccola: Added on last commit. timestamps will map to `string | undefined` so I had to add a check in the `decode-conversions` code which I _think_ makes sense

tomasciccola:
> This doesn't support the case of forwards-compatibility. E.g. the version of comapeo we release today needs to know which future versions of the file format it is able to parse. I suggest we use [comver](https://gitlab.com/staltz/comver) for the file format (`MAJOR.MINOR`). The value stored in this field will not be parsed (it's just there to support debugging in the future), but for the file reader we should attempt to parse this and throw if it can't be parsed e.g.

Ok. So `fileVersion` would be a string (and we can add a regex on the JSONSChema to validate the field when encoding). 
That validation pseudo-code you pasted would live on the `config-import` in `mapeo-core-next` right?

gmaclennan:
> Ok. So `fileVersion` would be a string (and we can add a regex on the JSONSChema to validate the field when encoding).

Yes, although I don't think regex in JSONSchema is necessary - this field will never be parsed by anything, it's just for info, so can be just left as an opaque string.

> That validation pseudo-code you pasted would live on the `config-import` in `mapeo-core-next` right?

Yes. The version name read from the config file does need to be parsable and conform to MAJOR.MINOR

proto/projectSettings/v1.proto
```diff
@@ -20,7 +20,14 @@ message ProjectSettings_1 {
     repeated bytes relation = 5;
   };
 
-  optional DefaultPresets defaultPresets = 2;
+  message ConfigMetadata {
+    string name = 1;
+    google.protobuf.Timestamp buildDate = 2;
+    google.protobuf.Timestamp importDate = 3;
+    string fileVersion = 4;
```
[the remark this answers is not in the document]
gmaclennan: Storing this info in the project settings is just for debug purposes - it will not need to be parsed. I think there is no need for the schema to be opinionated about it.

src/lib/decode-conversions.ts
```diff
@@ -50,9 +53,25 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
...
       : undefined,
+    configMetadata,
   }
 }
 
+function convertConfigMetadata(
+  configMetadata: ProjectSettings_1_ConfigMetadata | undefined
+): ProjectSettings['configMetadata'] {
+  if (!configMetadata) {
+    throw new Error('Missing required property configMetadata')
```
gmaclennan, line 64: This shouldn't be a required prop - a project could be created without any config imported yet.

src/lib/decode-conversions.ts
```diff
@@ -50,9 +53,25 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
...
+  }
+  if (!configMetadata.importDate) {
+    throw new Error('Missing required property configMetadata.importDate')
+  }
+  if (!configMetadata.buildDate) {
+    throw new Error('Missing required property configMetadata.buildDate')
+  }
+  return configMetadata as ProjectSettings['configMetadata']
```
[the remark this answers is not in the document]
tomasciccola: This checks are because `dates` in protobuf always generate types with `| undefined` but the rest of the fields are mandatory; so if we don't have type errors then we have the guarantee that when decoding, if `configMetadata` exists, then all fields _should_ be there. 
When encoding this shouldn't be an issue because types for JSON doesn't have this problems: if you set a date as required it won't be '| undefined'.
So in a sense, those checks only exist to please typescript

proto/projectSettings/v1.proto
```diff
@@ -20,7 +20,14 @@ message ProjectSettings_1 {
...
+  }
 
-  optional string name = 5;
+  optional DefaultPresets defaultPresets = 2;
+  optional ConfigMetadata configMetadata = 3;
```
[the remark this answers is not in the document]
tomasciccola: Fixed on last couple of commits

src/lib/decode-conversions.ts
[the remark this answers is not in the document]
gmaclennan: Not exactly mandatory. If `name` and `fileVersion` are missing from the protobuf, then they will be decoded as empty strings. The whole world of optional fields in protobuf is super confusing, including how they are decoded by ts-proto (see docs), but there are no easy solutions - it's a feature of protobuf that required fields are not supported. I'm not entirely sure what we should do - maybe we should leave everything optional and leave the API consumer to decide what to do with it if fields are missing, rather than throwing during decode?

gmaclennan (commented):
This has all left me confused about how we handle required/optional fields in protobuf. ts-proto gives the impression that primitives are "required", because the type is never `undefined`, but it actually decodes to defaults for primitives, like an empty string. I can't quite get my head around what we should do, maybe we could check in about this tomorrow?

tomasciccola:
> This has all left me confused about how we handle required/optional fields in protobuf. ts-proto gives the impression that primitives are "required", because the type is never `undefined`, but it actually decodes to defaults for primitives, like an empty string. I can't quite get my head around what we should do, maybe we could check in about this tomorrow?

Same, I thought I already knew were to deal with optionals but this threw me off. Same happened in #199. Basically every non-primitive type seems to be optional which is a drag. Basically I internally think about all this checks as "This should be impossible to happen but we need to please typescript" (but maybe is not impossible to happen?)

test/fixtures/good-docs-completed.js
```diff
@@ -88,6 +88,12 @@ export const goodDocsCompleted = [
         relation: cachedValues.defaultPresets.point,
       },
       name: 'myProject',
+      configMetadata: {
+        name: 'mapeo-config-1',
+        fileVersion: '1.0',
+        buildDate: cachedValues.configMetadata.buildDate,
+        importDate: cachedValues.configMetadata.importDate,
+      },
```
EvanHahn, line 96: Should this also have a config version?

EvanHahn (approved):

test/fixtures/good-docs-completed.js
tomasciccola: MHM, I mean, since we're not using the `configVersion` from metadata.json (for now) I didn't added to the schema. We can added later, but since we allow extra fields on the `metadataFile` (see the core-next PR) I see no issue with this. 
Just to clarify `configVersion` is the version of the built config, and `fileVersion` is the config format file version. But maybe `fileVersion` is not clear enough as a name??

EvanHahn (approved):
LGTM!

I think we can now merge without Gregor's approval.

PR #198, opened 2024-07-25
add unspecified field to enums

### LINKED ISSUE #171: Include an Unspecified Value in Enums
For each enum in our protobuf definitions, add `UNSPECIFIED` as the `0` enum, and increment the values for other enum values by `1`.

e.g.

```proto
enum AttachmentType {
    ATTACHMENT_TYPE_UNSPECIFIED = 0;
    photo = 1;
    video = 2;
    audio = 3;
}
```

We either need to map this to a default enum during decode, or we need to include this in the JSON schema / decoded type, and then add handling to the frontend.

tomasciccola:
should close #171. 
Instead of having a field "unspecified" for enums, I had to be deliberate (so `${enum_name_in_snake_case}_unspecified`) since protobufs don't allow enums to have fields with the same name

### DIFF
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -14,9 +14,10 @@ message DeviceInfo_1 {
   Common_1 common = 1;
 
   enum DeviceType {
-    mobile = 0;
-    tablet = 1;
-    desktop = 2;
+    device_type_unspecified = 0;
+    mobile = 1;
+    tablet = 2;
+    desktop = 3;
   }
 
   string name = 5;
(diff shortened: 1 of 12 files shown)


gmaclennan (commented):
Looks fine to me on a quick read through. I'll let Evan do the final sign-off.

tomasciccola:
@gmaclennan 

> Looks fine to me on a quick read through. I'll let Evan do the final sign-off.

I think you missed the comments on the issue -> https://github.com/digidem/mapeo-schema/issues/171#issuecomment-2252879592
I did a set of tests to see the case that you where talking about as justification for this work. 

I'm not against of merging this, just, want to understand what I'm missing ...

EvanHahn (approved):
Approving after a lot of back-and-forth on the issue. Sorry for the delay!

PR #199, opened 2024-07-25
feat: standarize refs across all docs

### LINKED ISSUE #195: Proposal: Standardize all refs to include both `docId` and `versionId`
**MVP Candidate**

This could potentially be included in MVP. It would be necessary for [storing presetIds on observations (#528)](https://github.com/digidem/comapeo-mobile/issues/528) to work correctly, because that would require field refs to also include versionIds.

**Is your feature request related to a problem? Please describe.**

We use refs (references) to link documents to other documents. Examples include:

- References to observations from a track
- References to fields from a preset
- References to nodes and ways (e.g. map features) from an observation (future)

Currently, a reference includes a `docId` and `docType`. This works, however it does not capture information about which _version_ of the referenced doc was linked. This is important for historical analysis of data, e.g. rolling back changes, or seeing what data was at a specific time, because the referenced documents could change, and we do not know what version was originally linked.

This is a problem for the proposal to [store presetIds on observations (#528)](https://github.com/digidem/comapeo-mobile/issues/528), because the preset versionId would reference the preset originally used to create the observation, but the `fieldIds` from the preset doc could not be used to identify which fields were actually used, because the `fieldId` only points to the most recent version of a field doc.

sub-problem: if we standardize the naming of reference fields, then we can optionally expand these references on the backend, e.g. return the doc with the reference(s) replaced with the actual document(s).

**Describe the solution you'd like**

We standardize all references to include both a `docId` and `versionId`. The `docType` of a reference is only necessary if the doc type is not implicit in the field, e.g. `fieldRef` would not need a doc type, because it implicitly only references docs of type field, but `refs` for a track could reference anything, although at the moment we are only supporting referencing observations.

For the initial implementation, we do not need to do anything with this information, but it is valuable to start storing it now so we can use it later. E.g. for now, we should resolve field references using the fieldRef docId, and ignore the fieldRef versionId, but we could use the fieldRef versionId at a later stage.

This would require changes to the code for tracks and for preset imports, to write the versionId as they are created or edited.

**Implementation tasks**

- [ ] Update schema to add `versionId` to refs
- [ ] Standardize naming of refs to `ref`, `refs` (array), or `${docType}Ref`, `${docType}Refs`
- [ ] Update preset import code in mapeo-core to also write field versionIds
- [ ] Update comapeo-mobile for new shape of field refs (change from `fieldIds: Array<string>` to `fieldRefs: Array<{ docId: string, versionId: string }>`)
- [ ] Update comapeo-mobile to write observation versionIds when creating a track.
- [ ] Check if comapeo-mobile needs any updates for reading referenced observations from a track.
- [ ] Check if there are any other places where a document references another document and needs to be standardized.

tomasciccola:
should close #195 
So, I tried to standarize references in docs by creating a `ref` protobuf and importing it in every document that needs it. 
For protobufs is useful that we can import, which I'm not doing in JSONSchema (I'm basically copy-pasting `ref` on documents that use it, changing only the description). This can became _kinda_ hard to track (but not a huge issue). I know JSONSchema allows to reference other schemas but I know that we're not using that feature for i.e. `Common` so there's probably a shortcoming I'm forgetting about.
Besides `docId` and `versionId` I added `type` so that we can track which type of documents are referenced. This was already used in tracks and translations (through `schemaRef`, now removed). Regarding this, I think it may make sense to have runtime checks (i.e. tracks should only reference observations) instead of having different `ref` types which only differ on the type of doc they reference


EvanHahn:
I'll review this soon. Could you fix the merge conflicts before I do?

tomasciccola:
Regarding asserting correct reference types:
1. Tracks should only reference observations
2. Presets should only reference fields or icons
3. Translations? (On the top of my head we translate fields and presets, but maybe we want to keep it lean?)

Maybe this is smth we don't need to do and we can be lean about it...

gmaclennan (commented):
Some initial feedback/thoughts on a quick read-through:

- A `versionId` is not a hex encoded string. We already have a proto def for this called [`Link`](https://github.com/digidem/mapeo-schema/blob/02a723576372313e102dad3ff45563efb7dbab05/proto/common/v1.proto#L10). We [encode it and decode it](https://github.com/digidem/mapeo-schema/blob/main/src/lib/utils.ts#L41-L60) by hex-encoding the discovery key and joining with `/` and the index.
- Also change `Link` to use the same definition.
- I think we should have a `ref` definition that does not include a `type`, with the type implicit. Otherwise we are allowing for invalid refs. E.g. a preset can have a `fieldRefs` field, which can only be refs to fields, and don't need a `type`. Right now the only place I think where we need a `type` for refs is for translations. Since it's not an array, maybe we can just keep it as a separate `refType` field? I'm not sure about this one and would welcome Evan's thoughts.
- Where possible, we should avoid calling this field `ref` or `refs`, and include a more descriptive name like `fieldRefs`, to make it easier to understand.
- For tracks, I think it's maybe ok to just say the refs are `observationRefs`. I think if we want to link other things to a track, we can do it as a separate field.

proto/translation/v1.proto
```diff
@@ -10,12 +11,11 @@ message Translation_1 {
...
   Common_1 common = 1;
+  Ref_1 ref = 2;
 
-  string schemaNameRef = 2;
-  bytes docIdRef = 3;
-  string fieldRef = 4;
-  string languageCode = 5;
-  string regionCode = 6;
-  string message = 7;
+  string fieldRef = 3;
```
EvanHahn, line 16: Should this be a `Ref_1`?

EvanHahn (changes requested):
> I think we should have a `ref` definition that does not include a `type`, with the type implicit. Otherwise we are allowing for invalid refs. E.g. a preset can have a `fieldRefs` field, which can only be refs to fields, and don't need a `type`. Right now the only place I think where we need a `type` for refs is for translations. Since it's not an array, maybe we can just keep it as a separate `refType` field? I'm not sure about this one and would welcome Evan's thoughts.

I agree with this. I think it makes more sense to have two fields: `fieldRefs` and `iconRefs`, or `observationRefs`.

That probably means removing `RefType` from `proto/ref/v1.proto`.

tomasciccola:
Following the feedback I:
* Add a `VersionId` proto to be used in `Common.links` and other refs
* Delete `Ref_1` proto
* Add specific `ref` messages to the corresponding protos (and schemas):
  1. `fieldRefs` and `iconRef` on presets
  2. `observationRefs` on tracks
  3. `docRef` on translation
* Add checks for undefined fields. This mostly happens when we have a list of refs (repeated) where, for some reason that I don't know, `versionId` can be `undefined`. Types are generated like that and from what I can see non-primitives in protobufs generate types with `| undefined`
* Fix types and tests
* I took de liberty of renaming `fieldRef` to `propertyRef` in translation since `fieldRef` can be misleading (its not related to `Field` whatsoever). `propertyRef` seemed more apt since its actually a 'dotprop' string. I can rollback that change if y'all feel like it

test/fixtures/bad-docs.js
```diff
@@ -82,4 +82,154 @@ export const badDocs = [
       color: '#ff00g1',
     },
   },
+  {
+    text: 'preset that references an invalid docTypr',
```
EvanHahn, line 86: Nitpick:

```suggestion
    text: 'preset that references an invalid docType',
```

test/fixtures/bad-docs.js
```diff
@@ -82,4 +82,154 @@ export const badDocs = [
...
+          type: 'preset',
+        },
+      ],
+      color: '#ff00ff',
+      terms: ['imastring'],
+      deleted: false,
+    },
+  },
+  {
+    text: 'track ref type can only refer to observations',
```
EvanHahn, line 181: When is this possible?

test/fixtures/bad-docs.js
```diff
@@ -82,4 +82,154 @@ export const badDocs = [
...
+      updatedAt: cachedValues.updatedAt,
+      createdBy: cachedValues.createdBy,
+      links: [],
+      name: 'myPreset',
+      geometry: ['point'],
+      tags: {},
+      addTags: {},
+      removeTags: {},
+      terms: [],
+      refs: [
```
EvanHahn, line 101: Is this possible?

test/fixtures/bad-docs.js
```diff
@@ -82,4 +82,154 @@ export const badDocs = [
...
+        someKeyForArrVal: ['arr', 'of', 'strings'],
+      },
+      addTags: {
+        heyAddThisTag: 'pleease',
+        orMaybeThis: ['right?', '', 'that was empty'],
+      },
+      removeTags: {
+        deleteInmeditaly: ['this list', 'of things'],
+      },
+      refs: [
```
EvanHahn, line 134: Is this possible?

EvanHahn (changes requested):
LGTM other than a few questions on tests.

tomasciccola:
> LGTM other than a few questions on tests.

Thanks for catching this. It was left from when `Ref` was the same across every docType and with a type field. 
Now it doesn't make sense to keep it

PR #207, opened 2024-08-05
feat: add `presetRef` to observation

### LINKED ISSUE #205: Add `presetRef: { docId, versionId }` to observation schema.

tomasciccola:
should close #205. I put it as required on JSONSchema and non-optional on the proto definition

### DIFF
--- a/src/lib/encode-conversions.ts
+++ b/src/lib/encode-conversions.ts
@@ -108,8 +108,8 @@ export const convertObservation: ConvertFunction<'observation'> = (
   let presetRef
   if (mapeoDoc.presetRef) {
     presetRef = {
-      docId: Buffer.from(mapeoDoc.presetRef?.docId, 'hex'),
-      versionId: parseVersionId(mapeoDoc.presetRef?.versionId),
+      docId: Buffer.from(mapeoDoc.presetRef.docId, 'hex'),
+      versionId: parseVersionId(mapeoDoc.presetRef.versionId),
     }
   }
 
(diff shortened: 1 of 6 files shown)


gmaclennan (changes requested):
I think we should leave `presetRef` as optional. I think there are cases where it might not make sense to populate this value (an observation could be created without a preset - it is in tests).

I think for Protobuf sub-messages, the difference between marking it as `optional` or not is in the generated code and resulting types either being `{ presetRef?: PresetRef | undefined }` or `{ presetRef: PresetRef | undefined }`. I think the former is preferable, to avoid the API consumer needing to always specify it as a property, even if undefined. So maybe add `optional` to the proto def too.

Can you ensure test fixtures include docs both with and without the presetRef set.

src/lib/encode-conversions.ts
```diff
@@ -105,13 +105,21 @@ export const convertObservation: ConvertFunction<'observation'> = (
   const metadata: Observation_1_Metadata = mapeoDoc.metadata && {
     ...Observation_1_Metadata.fromPartial(mapeoDoc.metadata),
   }
+  let presetRef
+  if (mapeoDoc.presetRef) {
+    presetRef = {
+      docId: Buffer.from(mapeoDoc.presetRef?.docId, 'hex'),
+      versionId: parseVersionId(mapeoDoc.presetRef?.versionId),
```
gmaclennan, line 112: The nullish check is unnecessary, and if it was it would result in a runtime error (`Buffer.from(undefined, 'hex')` throws an error).

```suggestion
      docId: Buffer.from(mapeoDoc.presetRef.docId, 'hex'),
      versionId: parseVersionId(mapeoDoc.presetRef.versionId),
```

gmaclennan (changes requested):
one small fix needed, once done go ahead and merge.

src/lib/encode-conversions.ts
tomasciccola: oh yeah, I forgot to remove it after adding the conditional. Thanks for the catch, merging now

PR #211, opened 2024-08-14
chore: move typedoc-related deps to dev deps

achou11:
Based on https://pkg-size.dev/@mapeo%2Fschema@3.0.0-next.19, I think including [`typedoc` is including a pinned version of typescript as a direct/peer dep](https://github.com/TypeStrong/typedoc/blob/v0.24.8/package.json), which causes installation for any module consuming @mapeo/schema to be incredibly slow (basically it's installing TS v5.1 alongside whatever TS version you've specified for yourself at the top level e.g. in an application project).

My understanding is that typedoc and any relevant plugins are only used at build-time, so it would make more sense for these to be defined as dev dependencies

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,9 +15,7 @@
         "compact-encoding": "^2.12.0",
         "glob": "^10.3.3",
         "protobufjs": "^7.2.5",
-        "type-fest": "^4.1.0",
-        "typedoc": "^0.24.8",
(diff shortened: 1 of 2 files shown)


EvanHahn (approved):
I see no reason these need to be production dependencies. Love it!

tomasciccola:
deaamnn

PR #212, opened 2024-08-14
chore: clean up deps

achou11:
Summary of changes:

- `@json-schema-tools/dereferencer`: upgrades to `^1.6.3` and move to dev deps
- `ajv`: move to dev deps
- `ajv-formats`: removed
- `glob`: update to v11 (removes node support for < 20) and move to dev deps

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "c8": "^8.0.1",
     "cpy-cli": "^5.0.0",
     "eslint": "^8.46.0",
-    "glob": "^11.0.0",
+    "glob": "^10.3.3",
     "hypercore": "^10.4.1",
     "json-schema-to-typescript": "^13.0.2",
     "mkdirp": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "@bufbuild/buf": "^1.26.1",
     "@json-schema-tools/dereferencer": "^1.6.3",
     "@types/json-schema": "^7.0.12",
+    "ajv": "^8.12.0",
     "c8": "^8.0.1",
     "cpy-cli": "^5.0.0",
     "eslint": "^8.46.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -34,10 +34,12 @@
   "homepage": "https://github.com/digidem/mapeo-schema#readme",
   "devDependencies": {
     "@bufbuild/buf": "^1.26.1",
+    "@json-schema-tools/dereferencer": "^1.6.3",
     "@types/json-schema": "^7.0.12",
     "c8": "^8.0.1",
     "cpy-cli": "^5.0.0",
     "eslint": "^8.46.0",
+    "glob": "^11.0.0",
```
EvanHahn, line 43: If this lacks support for Node <20 and we develop on Node 18, is this a safe update?

EvanHahn (changes requested):

package.json
achou11: hm right. i forgot that CI is doing this in Node 18 as well. will revert that change.

in terms of modules consuming this module, I wouldn't expect this to break if glob is defined as a dev dep

achou11: addressed via https://github.com/digidem/mapeo-schema/pull/212/commits/f4d96aa460230396330137ebcb317198c45b9ead

EvanHahn (approved):

package.json
achou11: fwiw it's worth though, CI on Node 18 passed before downgrading: https://github.com/digidem/mapeo-schema/pull/212/checks?sha=fe73a329800b2046875ff237217fdfcfbcdecab2

achou11:
just to confirm, is `ajv` truly only used in build time? This had me a little bit on the fence: 

https://github.com/digidem/mapeo-schema/blob/3a2ca671901414e7ae5cf913b4e771066596cf98/scripts/lib/generate-validations.js#L3 because i wasn't sure what that code pulls in. But based on the generated output, seems like it's adhering to "standalone" and not relying on the presence of `ajv` to be installed

EvanHahn:
It looks like that's only run by `npm run build`, which isn't run by consumers of this package (I think). I assume this is safe to do.

PR #213, opened 2024-08-14
feat: decode future schema versions without throwing

### LINKED ISSUE #168: Decode future schema versions without throwing
We should maintain the protobufs to be future-compatible, so we should not throw if the schema version does not match the one we are expecting. It would be good to do this for the MVP so that all public versions are future compatible.

EvanHahn:
We can now decode future schema versions without throwing. For example, if some future version encodes `Observation_2`, we can decode it as an `Observation_1`.

Closes #168.

### DIFF
--- a/scripts/lib/generate-config.js
+++ b/scripts/lib/generate-config.js
@@ -1,11 +1,7 @@
 /**
  * @param {ReturnType<import('./parse-config').parseConfig>} config
  */
-export function generateConfig({
-  currentSchemaVersions,
-  dataTypeIds,
-  protoTypeDefs,
-}) {
+export function generateConfig({ currentSchemaVersions, dataTypeIds }) {
   const idsLines = ['export const dataTypeIds = {']
   for (const [schemaName, dataTypeId] of Object.entries(dataTypeIds)) {
     idsLines.push(`  ${schemaName}: '${dataTypeId}',`)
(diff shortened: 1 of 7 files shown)


tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es2019",
-    "lib": ["es2020"],
+    "lib": ["es2022"],
```
EvanHahn, line 4: Necessary to use `Object.hasOwn`. Only affects type checking.

src/lib/utils.ts
```diff
@@ -14,12 +20,13 @@ export class ExhaustivenessError extends Error {
 
 /**
  * Get the name of the type, e.g. `Observation_5` for schemaName `observation`
- * and schemaVersion `1`
+ * and schemaVersion `5`
```
EvanHahn, line 23: This change isn't strictly related but I believe it's a typo.

scripts/lib/generate-config.js
```diff
@@ -20,26 +16,5 @@ export function generateConfig({
...
-    knownVersionsLines.push(`  ${schemaName}: [${schemaVersions.join(', ')}],`)
-  }
-  knownVersionsLines.push('}')
-
-  return [
-    ...idsLines,
-    '',
-    ...versionsLines,
-    '',
-    ...knownVersionsLines,
```
EvanHahn, line 42: `knownVersionsLines` is no longer used so we can remove it, and all its dependencies.

tomasciccola:
> if some future version encodes `Observation_2`, we can decode it as an `Observation_1`.

Is this actually reasoned like this? shouldn't we be able to decode `Observation_2` as `Observation_2` but with just newer fields ignored? 
This is just a semantics question though...

tomasciccola (approved):

gmaclennan:
Post-merge review: All looks good! As a non-urgent (and post-mvp) follow-up, maybe we can drop the schema version from being part of the protobuf name? I don't see a need for that now. But we can worry about this at a later time.

tomasciccola:
> Post-merge review: All looks good! As a non-urgent (and post-mvp) follow-up, maybe we can drop the schema version from being part of the protobuf name? I don't see a need for that now. But we can worry about this at a later time.

Yeah, having the schema version as part of the protobuf is not useful but I _think_ there was a reason for that? We should double check just in case

PR #216, opened 2024-08-19
feat: remove optional fields when unnecessary

### LINKED ISSUE #200: Remove `optional` fields from protobufs unless they're truly optional

tomasciccola:
should close #200


EvanHahn:
Lemme know when I should review this.

I could see splitting it into smaller PRs, but only if that's easier for you.

src/lib/decode-conversions.ts
```diff
@@ -95,7 +95,8 @@ export const convertObservation: ConvertFunction<'observation'> = (
     ...rest,
     attachments: message.attachments.map(convertAttachment),
     tags: convertTags(message.tags),
-    metadata: message.metadata || {},
+    // @ts-ignore
```
tomasciccola, line 98: Once I solve this type error, which I'm not being able for the time being, I think this is mostly ready to merge

EvanHahn (approved):
LGTM other than the `ts-ignore` error.

EvanHahn (approved):

src/lib/decode-conversions.ts
```diff
@@ -95,7 +95,7 @@ export const convertObservation: ConvertFunction<'observation'> = (
     ...rest,
     attachments: message.attachments.map(convertAttachment),
     tags: convertTags(message.tags),
-    metadata: message.metadata || {},
```
tomasciccola, line 98: If we set up a default value for metadata it should be smth like `{manualLocation:true}`, since `manualLocation` now is required. That field will default to `false` if undefined, so if we're okay with that, maybe we could leave it undefined?

EvanHahn: As long as we don't accidentally get `manualLocation: undefined` when _reading_, I'm fine with whatever.

EvanHahn (approved):

src/lib/decode-conversions.ts
tomasciccola: According to [this](https://protobuf.dev/programming-guides/proto3/#default) at least, we _should_ be fine

gmaclennan:
Just seeing this now following the issues with the front end. I think there are one or two things we need to review here. I can take a look Monday morning.

PR #224, opened 2024-08-28
feat!: remove `attachments` from tracks

### LINKED ISSUE #223: Remove `attachments` field from tracks schema
From an end-user perspective, it's not immediately clear as to why it exists in the first place. For now, it seems sufficient to reference attachments that are associated to the observations that are referenced by the `observationRefs` field.

tomasciccola:
should close #223

### DIFF
--- a/proto/track/v1.proto
+++ b/proto/track/v1.proto
@@ -16,27 +16,13 @@ message Track_1 {
   Common_1 common = 1;
   repeated Position locations = 2;
   repeated ObservationRef observationRefs = 3;
-  repeated Attachment attachments = 4;
   map<string, TagValue_1> tags = 5;
-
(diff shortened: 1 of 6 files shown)


proto/track/v1.proto
```diff
@@ -16,27 +16,13 @@ message Track_1 {
   Common_1 common = 1;
   repeated Position locations = 2;
   repeated ObservationRef observationRefs = 3;
-  repeated Attachment attachments = 4;
   map<string, TagValue_1> tags = 5;
```
EvanHahn, line 19: Could you change this to 4 now that we've removed 4?

EvanHahn (approved):
LGTM other than one small comment.

PR #229, opened 2024-08-29
feat!: make observation location optional

EvanHahn:
This is a partial revert of 4ce47f09ba980fbc70a978b1afc399e22313382e.

It turns out that observations can, in edge cases, lack a location. This makes its `lat` and `lon` fields optional.


ErikSin (approved):
Thanks!

PR #243, opened 2024-09-03
test: check code with ESLint

EvanHahn:
`npm run lint:eslint` (run as part of `npm lint` and `npm test`) now checks our code with ESLint.

This also fixes or disables all lint failures.


eslint.config.js
```diff
@@ -0,0 +1,31 @@
...
+  ...tseslint.configs.recommended,
+  {
+    ignores: ['dist/', 'generated/', 'intermediate/'],
+  },
+  {
+    rules: {
+      '@typescript-eslint/ban-ts-comment': [
+        'error',
+        {
+          'ts-expect-error': false,
```
EvanHahn, line 18: The default still allows this, but makes us leave a comment.

eslint.config.js
```diff
@@ -0,0 +1,31 @@
...
+        },
+      ],
+      '@typescript-eslint/no-explicit-any': 'off',
```
EvanHahn, line 21: We use a lot of `any`s so I don't think we should disable this right now.

eslint.config.js
```diff
@@ -0,0 +1,31 @@
...
+      '@typescript-eslint/no-unused-vars': [
+        'error',
+        {
+          varsIgnorePattern: '^_',
+          argsIgnorePattern: '^_',
+        },
+      ],
+    },
```
EvanHahn, line 29: Matches rules we have elsewhere.

src/lib/utils.ts
```diff
@@ -74,18 +74,16 @@ export function parseVersionId(versionId: string): VersionIdObject {
...
-    updatedAt,
-    deleted,
+    docId: _docId,
+    versionId: _versionId,
+    originalVersionId: _originalVersionId,
+    links: _links,
+    forks: _forks,
+    createdAt: _createdAt,
+    updatedAt: _updatedAt,
+    deleted: _deleted,
```
gmaclennan, line 85: I'm not a huge fan of re-writing code to satisfy a linter, although this may just be a personal preference.

When reviewing or writing code like this, it requires contextual knowledge of the eslint configuration to know why the code is written this way. Similarly a future developer might write variables prepended with a `_` and not be expecting eslint rules to be ignored because of it.

My preference is to use eslint ignore comments because it is more explicit about what the developer was intending.

test/fixtures/good-docs-completed.js
```diff
@@ -205,8 +205,7 @@ export const goodDocsCompleted = [
       updatedAt: cachedValues.updatedAt,
       links: [],
       name: 'my device name',
-      // @ts-expect-error
-      deviceType: 'motorbike',
+      deviceType: /** @type {any} */ ('motorbike'),
```
gmaclennan, line 208: Why change this? In the (extremely unlikely) scenario that we made `motorbike` a valid device type, we would want a @ts-expect-error to show an error here, because the test depends on this being a TS error.

src/lib/utils.ts
EvanHahn: Updated in eabe017905823df5bb5a5cd11cbbfa581d6a5cc3.

test/fixtures/good-docs-completed.js
EvanHahn: Removed in b1e28d4f9f9fd3a7ac176f6dcb5d04edc5191b2c.

gmaclennan (approved):
Looks good. Small niggle, can be ignored: we use [`filter-obj`](https://github.com/sindresorhus/filter-obj) in a couple of other modules, and I would lean towards using a well-maintained and tested module like that rather than our own implementation, which is doesn't have tests for edge-cases, e.g. symbol properties.

EvanHahn:
I'll plan do that in a followup.

PR #244, opened 2024-09-03
fix: `validate` shouldn't throw

EvanHahn:
`validate`'s interface was inconsistent. Sometimes, it would `throw` an error; sometimes, it would return `false`. This makes it consistent: always return a boolean.

I discovered this when I noticed that a test wasn't passing the right arguments to `validate`. When I fixed it, I got test failures.

(You could argue that this should work the other way: it should throw instead of returning a boolean. But that's a bigger breaking change.)

### DIFF
--- a/scripts/lib/generate-jsonschema-exports.js
+++ b/scripts/lib/generate-jsonschema-exports.js
@@ -2,10 +2,7 @@
 import { default as _ } from '@json-schema-tools/dereferencer'
 
 // Dereferencer's exports are all wrong for ESM imports
-/** @type {any} */
-const JsonSchemaDereferencer =
-  // @ts-ignore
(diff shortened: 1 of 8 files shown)


gmaclennan (changes requested):
hmm... I actually think throwing when you pass invalid function parameters (e.g. parameters that do not type-check) is ok behaviour. It's not actually validating anything in this case, so an error is expected. We don't generally do assertions in addition to type checks, but I'm ok with adding adding a more explicit error here like "invalid schema name". I think maybe it's the test that needs updating, with two flows: one for a valid schema name (returns false), and one for an invalid schema name (throws)

EvanHahn:
Closing in favor of #249.

PR #247, opened 2024-09-03
chore: remove `any`s from decode helper function

EvanHahn:
This change should have no user impact.

It's generally good to avoid `any`s, as they worsen type safety. This removes 3 of them in a `decode` helper function.


gmaclennan (approved):
fine... but this doesn't change type safety, it's just a syntax change.

PR #248, opened 2024-09-04
fix: observation metadata types

gmaclennan:
boy protobuf and typescript are tricky beasts...

So there were a few edge-cases around the way we type, encode and decode observation position metadata.

We were using `fromPartial` for encoding metadata, which assigns defaults for missing props, which meant that decoding would pass type checks, but defaults would be assigned to the missing props, in this case `longitude` and `latitude`, which is a "null island" bug.

This PR makes a few changes:

- `manualLocation` is now optional in protobuf - the distinction between when this is `undefined` (when there is no location for the observation) and when it is boolean (when there is a location) is important.
- `gpsAvailable`, `networkAvailable`, `passiveAvailable` props of `metadata.positionProvider` are now optional on the protobuf, to distinguish between cases when this data is not available on the client (iOS) and when it is.
- `locationServicesEnabled` of `metadata.positionProvider` is now required on the JSONSchema - if we do store position provider metadata, then we have this data - this corresponds to the Expo types.
- `timestamp` and `coords` are now required fields of `position` and `lastSavedPosition` in the JSON schema. These metadata options don't make sense without this data.
- If a protobuf is missing `timestamp` or `coords` (possible, because of the way protobuf make everything optional), then we strip the position prop from metadata (rather than the alternative of assigning defaults or throwing)

This PR also removes usage of `fromPartial`, and stops generating it since it's no longer needed. Using this is "dangerous" because it assigns default values deeply to a message, which can lead to unexpected results - data that is type safe, but might have required fields set to default values.

**Impact on frontend**:

This is breaking, because it changes types to make some fields required. However fields that are newly set to required are also required / always set on the Expo Location [`LocationObject`](https://docs.expo.dev/versions/latest/sdk/location/#locationobject) and [`LocationProviderStatus`](https://docs.expo.dev/versions/latest/sdk/location/#locationproviderstatus). Therefore in frontend code this should not break things, because these values are being set from these objects from Expo.

These changes are not really specific to Expo - it would not make sense to add this metadata without these required fields being available, even if we move away from Expo for reading location.

**Why is this important?**

In the future, this (position) metadata is potentially important for "proofs" of observations. Without these changes it is possible that default values could be unexpectedly set, e.g. values could be `false` when they were actually unknown (`undefined`).

### DIFF
--- a/buf.gen.yaml
+++ b/buf.gen.yaml
@@ -7,6 +7,8 @@ plugins:
     opt:
       - esModuleInterop=true
       - importSuffix=.js
+      - outputJsonMethods=false
+      - outputPartialMethods=false
       - env=node
       - snakeToCamel=false
       - useDate=string
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -482,10 +482,7 @@ function removeInvalidPositionMetadata(
 function removeInvalidPosition(
   position: Observation_1_Metadata_Position
 ): Position | undefined {
-  if (
-    typeof position.coords === 'undefined' ||
-    typeof position.timestamp === 'undefined'
-  ) {
+  if (position.coords === undefined || position.timestamp === undefined) {
     return undefined
   }
   return position as Position
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -100,7 +100,6 @@ export const convertObservation: ConvertFunction<'observation'> = (
     ...rest,
     attachments: message.attachments.map(convertAttachment),
     tags: convertTags(message.tags),
-    // Remove invalid position metadata if it's missing required fields
     metadata: metadata ? removeInvalidPositionMetadata(metadata) : {},
     presetRef,
   }
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -22,10 +22,15 @@ import {
 } from '../types.js'
 import { ExhaustivenessError, VersionIdObject, getVersionId } from './utils.js'
 import type { Observation, Track } from '../index.js'
-import type { Observation_1_Attachment } from '../proto/observation/v1.js'
+import type {
+  Observation_1_Attachment,
+  Observation_1_Metadata,
+  Observation_1_Metadata_Position,
+} from '../proto/observation/v1.js'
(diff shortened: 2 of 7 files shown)


src/lib/decode-conversions.ts
```diff
@@ -453,3 +459,35 @@ function convertTrackPosition(
...
+  }
+}
+
+function removeInvalidPosition(
+  position: Observation_1_Metadata_Position
+): Position | undefined {
+  if (
+    typeof position.coords === 'undefined' ||
+    typeof position.timestamp === 'undefined'
+  ) {
```
EvanHahn, line 489: Minor nitpick:

```suggestion
  if (position.coords === undefined || position.timestamp === undefined) {
```

src/lib/decode-conversions.ts
```diff
@@ -95,7 +100,8 @@ export const convertObservation: ConvertFunction<'observation'> = (
     ...rest,
     attachments: message.attachments.map(convertAttachment),
     tags: convertTags(message.tags),
-    metadata: message.metadata,
+    // Remove invalid position metadata if it's missing required fields
```
EvanHahn, line 103: Nit: I think the code is clear without this comment.

```suggestion
```

buf.gen.yaml
```diff
@@ -7,6 +7,8 @@ plugins:
     opt:
       - esModuleInterop=true
       - importSuffix=.js
+      - outputJsonMethods=false
```
EvanHahn, line 10: I think this is (1) a useful change (2) unrelated to the fix here. Totally fine to include it, but let me know if I misunderstand and this is actually important.

EvanHahn (approved):

buf.gen.yaml
gmaclennan: ah yes, it's unrelated to hear, but was related to me cleaning up unused methods, because I need to also mock them in comapeo-geometry. Should be in a different PR, but I think fine to leave them in here.

src/lib/decode-conversions.ts
gmaclennan: This is an old habit from pre-ES5 days: https://stackoverflow.com/questions/8783510/how-dangerous-is-it-in-javascript-really-to-assume-undefined-is-not-overwritte

PR #253, opened 2024-09-10
fix!: remove `isInitialProject` from project settings

EvanHahn:
[Gregor rightly pointed out that this value is synced][0]. Given that this value is currently unused on the frontend, we should remove it.

[0]: https://github.com/digidem/comapeo-core/issues/751#issuecomment-2337594457


gmaclennan (approved):
The front end isn't using this right? So no impact from deleting.

PR #254, opened 2024-09-11
chore!: improve decoding of required fields

### LINKED ISSUE #252: A new take on protobuf optional fields...
I was reviewing some code here today and remembered why we originally had some required fields as `optional` on the proto definitions.

TL;DR I think any _required_ field that is a boolean or number type should be `optional` in the proto definition, but we could choose to just ignore this issue.

Why? Because all fields are optional in proto3, and a missing field is decoded as the default value (`0` for a number type, `false` for a boolean). Because the default values of number and boolean types has a specific meaning, there is no way to distinguish when decoding between a "missing" value (e.g. the protobuf was encoded without the value) and an actual value.

e.g. given this protobuf definition:

```proto
message Test {
  bool field1 = 1;
}
```

If I decode an empty arraybuffer `Test.decode(new Uint8Array())` then this will decode to `{ field1: false }`.

However if I use this protobuf definition:

```proto
message Test {
  optional bool field1 = 1;
}
```

Then decoding an empty arraybuffer will result in `{ field1: undefined }`

If we are treating some fields as "required", e.g. they should always be set by the client, then if they are not marked as `optional`, they could actually be missing, and we could unintentionally get the default value back.

This is ok for `bytes` and `string` field types, because we can check for an empty Buffer or empty string if they are actually required.

This is an issue though for the number and boolean types. Let's say we have required field for `longitude`. If we have a protobuf definition:

```proto
messate Test {
  optional double longitude = 1;
}
```

then if we decode a buffer that is missing this field, we will get back an object with `longitude: 0`, and would be unable to distinguish it from a message where `longitude` is intentionally `0`.

## So how do I actually get required fields?

There is no reliable way to "enforce" required fields within the decoder. The default behaviour in standard protobuf decoding to use the default value is probably not what we want for most required fields use-cases. Instead we need to have explicit actions in our own decode code to detect when they are missing (which confusingly requires us to mark them as `optional` on the protobuf), and then take the appropriate action, which may be to throw an error (e.g. just not decode / ignore the message), or could be to set a default value, or could be to remove a property that has no meaning without the required values.

## Breaking change?

Adding `optional` to a number or boolean field will not break decoding (as in, messages will be decoded without error), but the decoded value will change. In the example above, if we start with the non-optional boolean `field1` and we have a message that was encoded from `{ field1: false }`, then if we change the proto definition to set that field as optional, then these existing messages will decode as `{ field1: undefined }`.

## Do we even need to worry about this?

Our types enforce fields being required at the API level, so there _should not_ be any cases where a required field is missing unintentionally. Our `encode()` functions would throw an error if we actually did not pass a required field to `encode()`. The only case where it is an issue is if buggy code had encoded a message, or we have an old message type without the field.

So... We probably don't need to worry about this, but if we do actually want to be strict about "required" fields, then we should mark them as optional, check for them in our own decode / convert functions, and throw if they are missing.

EvanHahn:
A required field, confusingly, should probably be marked *optional* in the protobuf. That way, we can throw errors when decoding. See [#252] for more details.

This change:

- Updates `.proto` files to mark all required fields with our custom `required` tag
- Updates `.proto` files to mark most required fields as `optional`, so we can throw errors when decoding
- Throw errors when decoding if a field is missing

I tested this end-to-end.

Closes #252.

[#252]: https://github.com/digidem/mapeo-schema/issues/252

### DIFF
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -21,6 +21,6 @@ message DeviceInfo_1 {
     selfHostedServer = 4;
   }
 
-  string name = 5 [(required) = true];
+  string name = 5;
   DeviceType deviceType = 6;
 }
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -21,6 +21,6 @@ message DeviceInfo_1 {
     selfHostedServer = 4;
   }
 
-  string name = 5;
+  optional string name = 5 [(required) = true];
   optional DeviceType deviceType = 6;
 }
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -22,5 +22,5 @@ message DeviceInfo_1 {
   }
 
   string name = 5 [(required) = true];
-  optional DeviceType deviceType = 6;
+  DeviceType deviceType = 6;
 }
--- a/proto/deviceInfo/v1.proto
+++ b/proto/deviceInfo/v1.proto
@@ -21,6 +21,6 @@ message DeviceInfo_1 {
     selfHostedServer = 4;
   }
 
-  optional string name = 5 [(required) = true];
+  string name = 5 [(required) = true];
   optional DeviceType deviceType = 6;
 }
--- a/proto/preset/v1.proto
+++ b/proto/preset/v1.proto
@@ -23,7 +23,7 @@ message Preset_1 {
   repeated FieldRef fieldRefs = 10;
   optional IconRef iconRef = 11;
   repeated string terms = 12;
-  string color = 13 [(required) = true];
+  optional string color = 13;
 
   enum Geometry {
     geometry_unspecified = 0;
--- a/proto/preset/v1.proto
+++ b/proto/preset/v1.proto
@@ -15,7 +15,7 @@ message Preset_1 {
 
   Common_1 common = 1;
 
-  optional string name = 5 [(required) = true];
+  string name = 5 [(required) = true];
   repeated Geometry geometry = 6;
   map<string, TagValue_1> tags = 7;
   map<string, TagValue_1> addTags = 8;
@@ -23,7 +23,7 @@ message Preset_1 {
   repeated FieldRef fieldRefs = 10;
   optional IconRef iconRef = 11;
   repeated string terms = 12;
-  optional string color = 13 [(required) = true];
+  string color = 13 [(required) = true];
 
   enum Geometry {
     geometry_unspecified = 0;
--- a/proto/preset/v1.proto
+++ b/proto/preset/v1.proto
@@ -15,7 +15,7 @@ message Preset_1 {
 
   Common_1 common = 1;
 
-  string name = 5 [(required) = true];
+  string name = 5;
   repeated Geometry geometry = 6;
   map<string, TagValue_1> tags = 7;
   map<string, TagValue_1> addTags = 8;
--- a/proto/preset/v1.proto
+++ b/proto/preset/v1.proto
@@ -15,7 +15,7 @@ message Preset_1 {
 
   Common_1 common = 1;
 
-  string name = 5;
+  optional string name = 5 [(required) = true];
   repeated Geometry geometry = 6;
   map<string, TagValue_1> tags = 7;
   map<string, TagValue_1> addTags = 8;
@@ -23,7 +23,7 @@ message Preset_1 {
   repeated FieldRef fieldRefs = 10;
   optional IconRef iconRef = 11;
   repeated string terms = 12;
-  string color = 13;
+  optional string color = 13 [(required) = true];
 
   enum Geometry {
     geometry_unspecified = 0;
--- a/schema/deviceInfo/v1.json
+++ b/schema/deviceInfo/v1.json
@@ -10,8 +10,7 @@
     },
     "name": {
       "type": "string",
-      "description": "Name of the device",
-      "minLength": 1
+      "description": "Name of the device"
     },
     "deviceType": {
       "type": "string",
@@ -26,6 +25,6 @@
       "description": "Type of device"
     }
   },
-  "required": ["schemaName", "name"],
+  "required": ["schemaName"],
   "additionalProperties": false
 }
--- a/schema/deviceInfo/v1.json
+++ b/schema/deviceInfo/v1.json
@@ -10,7 +10,8 @@
     },
     "name": {
       "type": "string",
-      "description": "Name of the device"
+      "description": "Name of the device",
+      "minLength": 1
     },
     "deviceType": {
       "type": "string",
--- a/schema/deviceInfo/v1.json
+++ b/schema/deviceInfo/v1.json
@@ -25,6 +25,6 @@
       "description": "Type of device"
     }
   },
-  "required": ["schemaName", "deviceType"],
+  "required": ["schemaName", "name", "deviceType"],
   "additionalProperties": false
 }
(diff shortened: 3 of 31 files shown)


gmaclennan:
I have a question about enums: without `optional` then if the field is not present, it will be decoded to unspecified. With optional we kind of have two "unspecified" states of the enum. Because we have already gone to the effort of adding unspecified everywhere, I think maybe we could follow a practice of enum fields never being optional? And require their presence base on unspecified.

EvanHahn:
> I have a question about enums: without `optional` then if the field is not present, it will be decoded to unspecified. With optional we kind of have two "unspecified" states of the enum. Because we have already gone to the effort of adding unspecified everywhere, I think maybe we could follow a practice of enum fields never being optional? And require their presence base on unspecified.

Fixed the one violation of this in 931f47383377a04a3a6d74e6dd879516a709ffd. (Manually checked every occurrence of `git grep -E "optional [A-Z]" proto/`.)

> I realise there is confusing behaviour here: you could write data with an empty string for a required field and it will type check, and it will not complain when writing to the DB, you will just never see that data record again...

Good point. Gave all `required` strings a `minLength` in 0bd5ab41f3c1cba2fe199cbdd5884130af336228. (Required a small tweak to the way we do codegen because of a bug in Ajv.)

EvanHahn:
This causes breakages in `@comapeo/core` which I'll look at tomorrow.

proto/preset/v1.proto
```diff
@@ -15,15 +15,15 @@ message Preset_1 {
...
+  string name = 5 [(required) = true];
   repeated Geometry geometry = 6;
   map<string, TagValue_1> tags = 7;
   map<string, TagValue_1> addTags = 8;
   map<string, TagValue_1> removeTags = 9;
   repeated FieldRef fieldRefs = 10;
   optional IconRef iconRef = 11;
   repeated string terms = 12;
-  string color = 13;
+  string color = 13 [(required) = true];
```
gmaclennan, line 26: I think maybe not make this required. It will need a modification to the frontend to handle the undefined case and use a default colour.

proto/preset/v1.proto
```diff
@@ -15,15 +15,15 @@ message Preset_1 {
 
   Common_1 common = 1;
 
-  string name = 5;
+  string name = 5 [(required) = true];
```
gmaclennan, line 18: I think maybe not make this required. EIther frontend or backend can fallback to a default name.

proto/deviceInfo/v1.proto
```diff
@@ -21,6 +21,6 @@ message DeviceInfo_1 {
     selfHostedServer = 4;
   }
 
-  string name = 5;
-  optional DeviceType deviceType = 6;
+  string name = 5 [(required) = true];
```
gmaclennan, line 24: Like other cases, maybe don't make this required. The frontend can use an empty string, or the device ID, if it is not set.

src/lib/decode-conversions.ts
```diff
@@ -86,9 +97,7 @@ export const convertObservation: ConvertFunction<'observation'> = (
   let presetRef
 
   if (rest.presetRef) {
-    if (!rest.presetRef.versionId)
-      throw new Error('found presetRef on observation but is missing versionId')
-
+    ensure(rest.presetRef.versionId, 'observation.presetRef', 'versionId')
     presetRef = {
       docId: rest.presetRef.docId.toString('hex'),
       versionId: getVersionId(rest.presetRef.versionId),
```
gmaclennan, line 103: getVersionId should be modified to use the ensure function

src/lib/decode-conversions.ts
```diff
@@ -494,8 +528,6 @@ function removeInvalidPositionMetadata(
 function removeInvalidPosition(
   position: Observation_1_Metadata_Position
 ): Position | undefined {
-  if (position.coords === undefined || position.timestamp === undefined) {
-    return undefined
-  }
-  return position as Position
+  const { timestamp, coords, ...rest } = position
+  return coords && timestamp ? { timestamp, coords, ...rest } : undefined
 }
```
gmaclennan, line 533: I don't like this change because 
1. I find it harder to read
2. It creates a 2 new objects for every decode (`rest` and the returned value)

schema/deviceInfo/v1.json
```diff
@@ -10,7 +10,8 @@
     },
     "name": {
       "type": "string",
-      "description": "Name of the device"
+      "description": "Name of the device",
+      "minLength": 1
     },
     "deviceType": {
```
gmaclennan, line 16: Thinking this should actually be required in the JSON Schema

gmaclennan (commented):
Did an initial review of this because I'm not sure when I can get back to it later.

I think the barrier for a "required" field at the protobuf level should be a bit higher. I think it should be "the app can't function without this field", because the action we take is fairly "drastic" which is to ignore the record altogether. There are a few fields I have highlighted that I think could be made optional. I think also the config metadata should also maybe be all optional - I feel like in general metadata should be a "optional if you have it" field.

VersionId should have `index` as required, and decode code should check that. I think we we require it on the JSON side.

There are a few follow up things I think which are not essential for MVP, such as checking docIds are not zero-length, and I think we should also look at distinguishing "write" vs "read" types for the app-side, mainly for enums - you should not be able to write an enum which is "unspecified" or "unrecognized", but the read type shoudl have them.

proto/preset/v1.proto
EvanHahn: In 1bfedae88357107e73184a2328313242875faa59, I:

- removed the `required` modifier from the proto
- removed the `minLength` requirement from the JSON Schema
- made it optional in the JSON Schema

proto/deviceInfo/v1.proto
EvanHahn: In 1bfedae88357107e73184a2328313242875faa59, I:

- removed the `required` modifier from the proto
- removed the `minLength` requirement from the JSON Schema
- made it optional in the JSON Schema

src/lib/decode-conversions.ts
EvanHahn: I did this to avoid a type cast, but I don't think this change should have been included in this PR anyway. Reverted in b66076da7cbbb247aa3300949d6db68a67b15a68.

schema/deviceInfo/v1.json
EvanHahn: Made it required in 3beea8732763ec688d620dcec146fbeb7083d15e.

proto/preset/v1.proto
tomasciccola: unless someone builds a config without the config builder, color shouldn't be missing. If a preset doesn't have a color on a specific config, it gets filled with a default color when building a config. So a preset _should always_ have a color 
(I dunno if this comments helps or not...)

src/lib/decode-conversions.ts
EvanHahn: Done in a3b10cedb5785fd8d0964c6396300910c3829e90.

proto/preset/v1.proto
EvanHahn: I believe I addressed this in e585e2dad51c1ba75cc551fc1e598d23f7e4d50c.

- `color` is optional in both the proto and the JSON Schema
- if defined, it must match the hex color regex

This should be impossible because people should be using the settings builder, but it could theoretically happen and this now stops it.

EvanHahn:
I _think_ I got this working, but I'm going to test it end-to-end to make sure things still work.

EvanHahn:
Running into issues getting this to work in `@comapeo/core` and beyond. I'll look more on Monday.

EvanHahn:
This PR was getting a little too large for me, so I'm starting to split it into smaller ones:

- #257
- #258

EvanHahn:
Gregor and I paired on this so I'm going to merge.

PR #255, opened 2024-09-11
fix!: property decode new `oneof` cases

EvanHahn:
We have proto definitions that use `oneof`, like this:

```protobuf
message Foo {
  oneof something {
    string bar = 1;
    string baz = 2;
  }
}
```

We also have decoding that does something like this:

```javascript
switch (foo.something?.$case) {
  case 'bar':
    // ...do something...
  case 'baz':
    // ...do something...
  default:
    throw new Error('Unexpected case')
}
```

This will break if we add another case, like this:

```diff
 message Foo {
   oneof something {
     string bar = 1;
     string baz = 2;
+    string qux = 3;
   }
 }
```

This fixes those cases in the codebase, treating an unknown `oneof` as skipped. We already did this in a few places.

I manually tested this (by adding a new `oneof` case) to make sure things would still decode correctly if an old version decoded something from a new version.

### DIFF
--- a/src/lib/decode-conversions.ts
+++ b/src/lib/decode-conversions.ts
@@ -21,7 +21,7 @@ import {
   type MapeoDocDecode,
 } from '../types.js'
 import { ExhaustivenessError, VersionIdObject, getVersionId } from './utils.js'
-import type { Observation, Track } from '../index.js'
+import type { Icon, Observation, Track } from '../index.js'
 import type {
   Observation_1_Attachment,
   Observation_1_Metadata,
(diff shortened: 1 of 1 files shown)


src/lib/decode-conversions.ts
```diff
@@ -384,8 +395,7 @@ function convertTagValue({ kind }: TagValue_1): JsonTagValue {
 function convertTagPrimitive({
   kind,
 }: TagValue_1_PrimitiveValue): TagValuePrimitive {
-  if (!kind) return undefined
-  switch (kind.$case) {
+  switch (kind?.$case) {
```
EvanHahn, line 398: Same comment as above.

src/lib/decode-conversions.ts
```diff
@@ -362,8 +372,7 @@ function convertTags(tags: { [key: string]: TagValue_1 } | undefined): {
 }
 
 function convertTagValue({ kind }: TagValue_1): JsonTagValue {
-  if (!kind) return undefined
-  switch (kind.$case) {
+  switch (kind?.$case) {
```
EvanHahn, line 375: I don't think this changes functionality at all, but (1) I think it's a little more readable (2) it handles the case where `kind` is defined but `$case` isn't for some reason.

gmaclennan (approved):
Looks good, good catch.

PR #256, opened 2024-09-12
chore: work around Ajv module bug in generated code

EvanHahn:
JavaScript module inconsistency strikes again.

In some cases, Ajv will inject `require` calls into generated code even if you ask it to generate ESM. [This is a bug][0].

For example, here's something it adds if you use [the string `minLength` or `maxLength` properties][1]:

```javascript
// ...
const func2 = require("ajv/dist/runtime/ucs2length").default;
// ...
```

That won't work in an ESM environment.

As a workaround, I inject the following into our generated code:

```javascript
import { createRequire } from 'node:module';
const require = createRequire(import.meta.url);
```

This, along with moving `ajv` to production dependencies, should work around this issue until the bug is fixed.

*This isn't a problem today*, but will immediately become a problem if we start using one of those validators. It's a bug waiting to happen.

I think this is a useful change on its own, but it should hopefully make [an upcoming change easier](https://github.com/digidem/comapeo-schema/pull/254).

[0]: https://github.com/ajv-validator/ajv/issues/2209
[1]: https://json-schema.org/understanding-json-schema/reference/string#length

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "3.0.0-next.26",
       "license": "MIT",
       "dependencies": {
+        "ajv": "^8.12.0",
         "compact-encoding": "^2.12.0",
         "protobufjs": "^7.2.5",
         "type-fest": "^4.26.0"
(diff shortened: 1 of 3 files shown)


tomasciccola (approved):
deamn, nice catch!

PR #264, opened 2024-09-17
chore: make ajv a dev dep

gmaclennan:
Now that we vendor ucs2length, ajv is not needed at runtime


gmaclennan (commented):
ajv at runtime was causing react native to crash

EvanHahn (approved):

PR #277, opened 2024-12-03
feat: omitServerProps helper function

gmaclennan:
I needed a more flexible version of `valueOf` for removing server-managed props from a Mapeo Doc. In the front-end we are modifying some props such as replacing refs with actual docs, but we still need to be able to strip server properties when editing.

I decided to create a separate function since this is subtly different from `valueOf`, and the name `valueOf` isn't great anyway.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,9 +1,12 @@
+import { excludeKeys } from 'filter-obj'
 import { type ProtoTypeNames } from '../proto/types.js'
 import {
+  type FilterBySchemaName,
+  type MapeoCommon,
   type MapeoDoc,
   type MapeoValue,
-  type FilterBySchemaName,
 } from '../types.js'
+import { commonSchema } from '../schemas.js'
 import { omit } from './omit.js'
 
 export function getOwn<T extends object, K extends keyof T>(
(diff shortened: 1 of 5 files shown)


src/lib/utils.ts
```diff
@@ -93,3 +100,22 @@ export function valueOf<TDoc extends MapeoDoc>(
...
+const keysToExclude = [...commonSchemaKeys, 'forks'] as const
+
+/**
+ * Omit server-managed properties from a document, such as `createdBy`,
+ * `createdAt`, etc. These properties cannot be modified by the client.
+ */
+export function omitServerProps<TDoc extends MapeoCommon>(
+  doc: TDoc & { forks?: string[] }
+) {
+  return excludeKeys(doc, keysToExclude)
```
EvanHahn, line 120: Can we use our existing `omit` helper (and maybe remove the new `filter-obj` dependency)?

src/lib/utils.ts
```diff
@@ -79,6 +82,10 @@ export function parseVersionId(versionId: string): VersionIdObject {
   return { coreDiscoveryKey, index }
 }
 
+/**
+ * Get the value of a document, excluding server-managed properties.
```
EvanHahn, line 86: I think this comment is a little unclear because it makes it look like it should do the same thing as the other function.

EvanHahn (approved):
1. I was initially surprised to see that the word "server" meant "managed by the backend"...maybe there's a better name for this?
1. I admit I don't fully understand the motivation for this change but I trust that y'all have discussed it and have a good reason.
1. Do we have a plan to deprecate `valueOf`?

PR #280, opened 2024-12-17
feat: add migration metadata to observation schema

EvanHahn:
This adds a new field, `migrationMetadata`, which has two sub-fields:
    
- `originalDocument`: the original observation, in case we ever need it
- `hypercore`: metadata about the observation in the old Hypercore

I considered adding this to the `Common` schema, but I don't think it makes sense there because we won't necessarily be migrating all types, like `CoreOwnership`.

### DIFF
--- a/proto/observation/v1.proto
+++ b/proto/observation/v1.proto
@@ -75,4 +75,16 @@ message Observation_1 {
     bytes docId = 1;
     VersionId_1 versionId = 2;
   }
+
+  message MigrationMetadata {
+    bytes originalDocument = 1;
+    message HypercoreMetadata {
+      optional string rootHashChecksum = 1;
+      optional string signature = 2;
+      optional string coreKey = 3;
+      optional uint32 blockIndex = 4;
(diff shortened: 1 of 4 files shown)


tomasciccola (approved):

gmaclennan:
I've reverted this because I think it contains bugs. It seems like the original observation is stored in the protobuf as a Buffer, in JSONSchema it is a hex-encoded buffer, but the decoding logic is trying to stringify the Buffer. This will lead to corrupt data when editing observations.

The "raw" format of migrated observations is JSON (this is what is stored in the MLEF), so I'm not sure where we would get the buffer from in the first place.

Rather than fix this, which would take a while, I want to revert this quickly so that we don't introduce these breaking changes in any build of comapeo-core.

I wonder whether we should store the original doc as JSON / struct on the protobuf. Also, I wonder whether for the case of migrated observation, we should not store it for edits, e.g. when `links: []` is not an empty array.

PR #285, opened 2025-04-15
feat: add project color and description fields to project settings

achou11:
Addresses https://github.com/digidem/comapeo-core/issues/1014

- `projectDescription` is an **optional** string
- `projectColor` is an **optional** string, where the string must be hex formatted (e.g. `#000000`). we could make this a required field but I believe that would require backfilling a default value for documents from older schema versions. I feel a bit uneasy about the schema being prescriptive for something like that, which is usually app-specific.

### DIFF
--- a/proto/projectSettings/v1.proto
+++ b/proto/projectSettings/v1.proto
@@ -30,4 +30,6 @@ message ProjectSettings_1 {
   optional DefaultPresets defaultPresets = 2;
   optional ConfigMetadata configMetadata = 3;
   optional string name = 4;
+  optional string projectDescription = 5;
+  optional string projectColor = 6;
 }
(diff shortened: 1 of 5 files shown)


achou11:
We don't have many examples of us adding a required field to a schema and how that affects migration (if at all). The more I'm reading around, the less certain I am about whether the `projectColor` should be optional or not. As the relevant issue mentions, ideally it would be required and if I make that bare minimum change (i.e. updating the json schema + proto file) there are no types issues that are raised (when I would've expected some in the decoding logic especially).

RangerMauve (approved):
This LGTM. I think the color validation is good as is since there's precedent in other fields

PR #289, opened 2025-04-30
feat: Exif fields for attachments main file to use in frontend view

RangerMauve:
For https://www.notion.so/digidem/Exposing-Observation-Data-GPS-Photos-6f9cd1cf9f0e40ed8eda96b6a70f3aa3#bf0c24e19a0b454194217f7055df1337

Based on [EXIF](https://exiftool.org/TagNames/EXIF.html) tag names we can expect from expo-camera. Also the [GPS](https://exiftool.org/TagNames/GPS.html) tag names.

Request for feedback:
- Any fields I should add or remove?
- Should I change the JSON schema to use `oneOf` to ensure only attachments of type Image can have exif properties, or is app level checking okay? If so how will it interplay with typescript?
- Any clue why the "good docs" test is failing when I add the exif tag?


achou11:
Hesitant about adding this information to the attachments schema - my thoughts on this align with what's mentioned in https://github.com/digidem/comapeo-core/issues/903#issuecomment-2428707408. EXIF data and similar will vary across the actual blobs. Attachments in this schema can/will point to various blobs, so I'm not sure it makes sense to have a single value that's supposed to be representative of all of the blobs that an attachment points to.

ErikSin:
> Hesitant about adding this information to the attachments schema - my thoughts on this align with what's mentioned in [digidem/comapeo-core#903 (comment)](https://github.com/digidem/comapeo-core/issues/903#issuecomment-2428707408). EXIF data and similar will vary across the actual blobs. Attachments in this schema can/will point to various blobs, so I'm not sure it makes sense to have a single value that's supposed to be representative of all of the blobs that an attachment points to.

The issue is that we need to access this EXIF data even when the original photo is not availble. When the user has there media manager set to only sync previews (which based on Mapeo-legacy will be most users), we still want to be able to present this data.

RangerMauve:
> Couldn't EXIF data and other metadata vary across the actual blobs?

It's useful to keep in mind that this feature is for the user story of viewing information about a specific photo. Even if the variants like previews are going to be different (or likely have the exif stripped) we want to tell the user what this data was for the original file that was uploaded.

achou11:
> Even if the variants like previews are going to be different (or likely have the exif stripped) we want to tell the user what this data was for the original file that was uploaded.

This helps my understanding of the proposed change a lot more. Though it feels like an important detail that should be documented or reflected in the naming of things introduced by this PR.

Maybe completely farfetched or unrelated: is it always assumed that the variants of a blob that an attachment points to are actually derived from the original? In core, there's nothing stopping us from creating a blob that gets referenced as an attachment like this:

```ts
BlobApi.create({ originalUri: '/dog.jpg', previewUri: '/cat.jpg', thumbnailUri: '/mouse.jpg' })
```

Based on the intended goal, the attachment metadata would be associated with `dog.jpg`, right?

RangerMauve:
>  is it always assumed that the variants of a blob that an attachment points to are actually derived from the original?

I assumed so but I'm not sure if that's enforceable. You're correct that the attachment EXIF data should be associated with `dog.jpg`.

One thing to note is that this change is not part of the blobs API and is just for the Observation.attachments field which gets created on the front end.

---

Kind of an aside but could you check out why the test is failing? It seems the protobuf isn't properly encoding the exif field but I'm not sure why not.

achou11:
> Should I change the JSON schema to use oneOf to ensure only attachments of type Image can have exif properties, or is app level checking okay? If so how will it interplay with typescript?

Would say that if it's straightforward to implement, then it'd be cool to have that in place. Don't think it's a huge issue if not though. If implemented, one would have to do one of the following in order to soundly use the `exif` field without TS complaining:

- check the attachment `type`
- check that the `exif` field exists

[example in TS playground](https://www.typescriptlang.org/play/?pretty=false#code/C4TwDgpgBAgswEMDGALAthAdsKBeAUFFAD5QDehRUokAXFAORgoD2wLDANJURAB4BLAGYB+ehSpUAYixZioAZ2AAnAZgDmPKAF9K2kuWrgI9BggCuAEwEcDDAG4DLEW7vzukLTEqgJ4ydCxgBXoYZWUEEAAeOERUDGwAPjwoAG1KMiM6RmY2V24iTJoTRkdnV3wAXQ8vHwQUvzjA7AVUgAZq-AB6LqgAFQBlKE80MAAbBDUFfE9vFjGIADoxlnUACgRF-mEASnce-qGBBSgUBDAwEHxhKA3F4rxcXBzWdgYd8kpZhXmllfXNtshHs3Ac+idjqdzpdrkJbgwgQwoGpfB8JMNar9lqs7kCQUA)

RangerMauve:
@achou11 I think it'd be easiest for me to do that at the comapeo-core level in the method signature 🤔

ErikSin:
Should we be explicit in the naming that the exif data correlates to the original photo? Maybe `exifOfOriginalPhoto`. I know that the concern is that we are attaching an exif field to an attachment, but the attachment is purely a reference to any number of things. So perhaps naming it to recognize that will (at least partially) address the concern?

RangerMauve:
Good point. Some video and audio formats have exif-like metadata too but those have different fields.

https://en.wikipedia.org/wiki/ExifTool#Reading_support

Could be good to split this stuff out. Maybe a slightly less verbose name though? like `photoExif`? I think since the attachment has one blob hash it's safe to assume that any metadata would be related to that blob

ErikSin:
If thats the case though, perhaps we should put the work into enforcing the discriminate union at the schema level? I don't know how much work goes into that...so if it is very difficult open to discussing it more.

Its just that if we are going to start including a bunch of optional fields (there is a plan to include video eventually) I think it would be useful to enforce how they are being used at the schema level so there is less room for error.

RangerMauve:
I'll give it a shot with the [JSON Schema oneOf](https://json-schema.org/understanding-json-schema/reference/combining#oneOf). Just not sure if the generated typescript types can play nice with it

RangerMauve:
@ErikSin How's this? Turns out `oneOf` works decently!

test/fixtures/good-docs-completed.js
achou11, line 1: is it worth adding a good observation doc that has a non-photo attachment? similarly, maybe could update the bad docs test to include an observation with an invalid variant of an attachment

achou11 (approved):
Can't really comment on what kind of exif fields to include but overall changes lgtm

ErikSin (approved):
In terms of the exif data, I think what you have is overkill for what we need. Since they are all optional and coming directly from expo, it will work for our purposes and future proof us for any further information needed

RangerMauve:
Pushed some more tests!

PR #290, opened 2025-04-30
chore(docs): Explain process for adding new fields

RangerMauve:
Just so we can have this for future reference :P

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ const myObs: Observation = {
 - Add your fields with descriptions
 - Update the corresponding protocol buffer schema inside `proto/{name}/v1.proto`
 - Add the new field to `src/lib/decode-conversions.js` and `src/lib/encode-conversions.js`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,16 @@ const myObs: Observation = {
 
 ## API
 
+## Adding new values
+
+- Find the JSON schema inside `schemas/{name}/v1.json`
+- Add your fields with descriptions
+- Find the corresponding protocol buffer schema inside `proto/{name}/v1.proto`
```
achou11, line 52: Such a nitpick 😄 

```suggestion
- Update the corresponding protocol buffer schema inside `proto/{name}/v1.proto`
```

achou11: should probably mention running `npm run build` (or the more specific commands) after doing this

achou11 (approved):
Always took me a moment to remember what to do so this is super helpful! Suggested one notable step to detail

PR #299, opened 2025-06-05
feat!: add "external" field to attachment definition

### LINKED ISSUE #298: Add field to observation attachments definition to indicate whether attachment was created from external source
In the user-facing applications, we're introducing the idea of data being "validated" by the application. Until recently, it wasn't super clear what that really meant in the context of media. The definition that we've seem to have landed on is that "validated by CoMapeo" means that the relevant media for an attachment was created via a CoMapeo application (e.g. using the mobile app).

As it currently stands, we don't really have a great way of confirming if something was actually taken by a CoMapeo application. The proposed idea is to introduce a field to the attachment definition to indicate that something _wasn't_ taken by the application. For example, an optional boolean field called `external`, where the omission of the field suggests that the media for the attachment was created via the application.

achou11:
Closes #298 

Adds an ~optional~ boolean field to an attachment called `external`, which indicates whether the associated media for an attachment was created outside of CoMapeo e.g. manually uploading an image/video/audio file that was created from another device, application that isn't CoMapeo, etc.

The user-facing intent of this is to determine if the attachment is "validated" by CoMapeo.

### DIFF
--- a/proto/observation/v1.proto
+++ b/proto/observation/v1.proto
@@ -51,7 +51,7 @@ message Observation_1 {
     optional PhotoExif photoExif = 5;
     optional google.protobuf.Timestamp createdAt = 6;
     optional Position position = 7;
-    optional bool external = 8;
+    bool external = 8;
   }
   repeated Attachment attachments = 7;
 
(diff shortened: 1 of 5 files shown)


RangerMauve (approved):
LGTM~

RangerMauve:
@gmaclennan Yeah that's a good idea. It'll potentially save the frontend an extra check.

Is it safe to assume that all attachments thus far were not external? If yes I think non optional is best 👍

achou11:
Cool, I'll just adjust to make the field non-optional

RangerMauve:
Yeah might as well mark it breaking.

gmaclennan (approved):

PR #304, opened 2025-08-19
feat: Add collectStats project settings

RangerMauve:
Part of https://github.com/digidem/comapeo-core/issues/1096


gmaclennan (changes requested):
Fine, although I would call the setting "shareStats" - I think we should collect (as in calculate and index) the stats. Or maybe it's "sendStats". Naming is difficult! Collecting to me sounds like the device is collecting the stats, but this is referring to Awana "collecting" the stats, by the project members sending/sharing them with us.

gmaclennan (approved):

PR #306, opened 2025-08-27
Feat/project stats nonoptional

RangerMauve:
Per https://github.com/digidem/comapeo-core/pull/1098#discussion_r2303837955

### DIFF
--- a/src/lib/encode-conversions.ts
+++ b/src/lib/encode-conversions.ts
@@ -37,7 +37,6 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
   return {
     common: convertCommon(mapeoDoc),
     ...mapeoDoc,
-    sendStats: mapeoDoc.sendStats ?? false,
     defaultPresets: defaultPresets
       ? {
           point: defaultPresets.point.map((p) => Buffer.from(p, 'hex')),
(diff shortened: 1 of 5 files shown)


src/lib/encode-conversions.ts
```diff
@@ -37,6 +37,7 @@ export const convertProjectSettings: ConvertFunction<'projectSettings'> = (
   return {
     common: convertCommon(mapeoDoc),
     ...mapeoDoc,
+    sendStats: mapeoDoc.sendStats ?? false,
```
gmaclennan, line 40: This isn't necessary - `mapeoDoc.sendStats` is already a boolean, e.g. it cannot be undefined. Protobuf v3 is kind of confusing with how `optional` works, because in protobuf all fields are actually "optional" - an empty buffer would parse correctly. Non-optional fields fall back to defaults when decoded: if a boolean field is missing from the protobuf, it's decoded as `false`. Optional fields have an explicit `undefined` state. In this case, sendStats will be false when the field is not included in the protobuf (e.g. if reading project settings encoded with the previous schema version)

```suggestion
```

gmaclennan (commented):
Just need to remove the edit to encode-conversions.ts then we can merge.

